PR #4, opened 2017-08-07
Add dtest for CASSANDRA-13737

adelapena:
Add test verifying that a schema propagation adding a view over a non existing table doesn't prevent a node from start


adelapena:
Committed as [959208749d70e5808aec144e87b73e90d56a7f91](https://github.com/apache/cassandra-dtest/commit/959208749d70e5808aec144e87b73e90d56a7f91)

PR #8, opened 2017-09-06
Update ccm to 2.8.4

ptnapoleon:
This comes with a fix for an infinite loop in node.start


mshuler:
+1 committed to upstream git-wip

PR #13, opened 2017-11-23
Install ccm from github branch

spodkowinski:
Recently we had some ccm related issues with our dtest suite. In this case, it would be convenient to be able to merge dtest fixes into a separate branch on the ccm side and have dtests pick up the updated version automatically. This already happens with the python driver and I don't think has been an issue so far. WDYT @mambocab , @ptnapoleon ?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,8 @@
 futures
 six
 -e git+https://github.com/datastax/python-driver.git@cassandra-test#egg=cassandra-driver
-ccm==3.1.0
+# Used ccm version is tracked by a ccm git branch. Please create a PR there for fixes or upgrades to new releases.
+-e git+https://github.com/pcmanus/ccm.git@cassandra-test#egg=ccm
 cql
 decorator
 docopt


ptnapoleon:
So, I am -0.5 on this. We used to do this. The problem is that ccm has other consumers than dtest, and worse, is both an app and a library. As a result, sometimes changes go into ccm which break dtests, despite our best efforts at testing ccm changes. When we switched to versioning ccm, and manually pulling in ccm versions, it made development in both systems much easier.

For the scope of this project, we obviously only care about difficulty in developing on dtest. Anecdotally, people vastly preferred manual changes to the ccm dependency, as that's much more bisectable when debugging failures. The recent upgrade failures on C* trunk are due to a ccm change. When debugging that, I was able to bisect it to precisely the commit that bumped the ccm version, which let me easily find the ccm commit that caused the issue. Then I released ccm 3.1.1 with the fix.

When we used to pull in ccm automatically, bisecting issues caused by ccm didn't work, and one would have to tie in the time at which CI jobs ran to the time at which ccm commits were made. It's much more unpleasant. I do know we pull in the python driver from a branch. However, that branch isn't pushed to with every new commit to the new driver. Rather, commits are made there in a batch when a new, tested version of the driver is released, or very explicitly go in tied with new C* features (think a cqlsh feature going in alongside a driver feature). The people developing the drivers have far more resources to guarantee the correctness of the driver than I have to do the same with ccm.

Now, I know that's a lot of negatives up-front, sorry about that. Possible points in favor that I am aware of are:
1. People don't test dtest when bumping the ccm version anyway, so why bother with the formality
2. We can increase the speed at which ccm fixes are used in dtest CI
3. It is too inconvenient to get committers to push changes to dtest deps.

I think (1) is throwing the baby out with the bathwater, and for (2), is a handful of hours to run CI such a huge deal? It is not like the OSS C* project relies on dtest fixes having a very fast SLA. (3) is definitely true.

On the ccm side, how would you want me, effectively the only maintainer right now, to manage this branch? Either it gets every commit, in which case let's just pull from master, or it gets batched updates which are equivalent to releases. In the latter option, that's just pushing the burden of testing dtest to me, rather than the dtest developers (yes that also includes me, I know). It's also identical to just pulling in releases on the dtest side, except for how those releases are pulled in. If you want a branch that selectively pulls in commits just for dtest, that will diverge from master, my answer as the ccm maintainer is no, I will not maintain a fork.

Again, just -0.5 on this. Frankly, I don't develop on dtest much, so if this is easier for you and the other people who do, I totally okay with supporting the branch on the ccm side if you let me know how you want that branch maintained.

spodkowinski:
The main motivation here was to be able to fix dtests just by merging a PR on the ccm test-branch. Currently what needs to be done for ccm fixes to become effective, is to cut a ccm release and afterwards upgrade the version the dtest's `requirements.txt`. That's not very efficient if you just want to fix a single dtest.

In this case, it would be better to un-couple dtests from regular ccm releases, both to get dtests results green again more quickly and not having to publish ccm releases that are addressing issues not relevant outside the context of dtests (e.g. https://github.com/pcmanus/ccm/commit/ff47a9f4339da2dacea16713bb2360b1b04b70c2). 

As for bisecting, we could also use a git commit instead of a branch in `requirements.txt` Or use both, regular releases that are known to be working and individual commits for bug fixes in between.

ptnapoleon:
To make sure I understand you, PRs from dtest maintainers for issues in dtest that arise from ccm would be made to this test-branch, which I would then merge into master at whatever pace was comfortable/correct for ccm? 

When would things be merged into the ccm test branch from master?

spodkowinski:
Yes, that was my initial idea. Merging master into the test-branch would also work the same way, by creating a PR for the merge commit. It's a bit unorthodox to do it that way, as effectively the test-branch would become a fork of the repo maintained by the dtest devs. But given the close relationship between the projects, it's worth to give it some thoughts.

ptnapoleon:
That works for me.

PR #15, opened 2018-01-02
Cassandra 7544

jasobrown:


rebuild_test.py
```diff
@@ -140,7 +140,9 @@ def resumable_rebuild_test(self):
         """
         self.ignore_log_patterns = list(self.ignore_log_patterns) + [r'Error while rebuilding node',
                                                                      r'Streaming error occurred on session with peer 127.0.0.3',
-                                                                     r'Remote peer 127.0.0.3 failed stream session']
+                                                                     r'Remote peer 127.0.0.3 failed stream session',
+                                                                     r'Streaming error occurred on session with peer 127.0.0.3:7000',
```
jasobrown: how will this work if we use a single (network) interface? or configurable ports? is it best to get the address of node3 and plug that into the patterns, or is that opening up a large can of worms?

repair_tests/incremental_repair_test.py
```diff
@@ -168,9 +171,10 @@ def _make_fake_session(self, keyspace, table):
         for node in self.cluster.nodelist():
             session = self.patient_exclusive_cql_connection(node)
             session.execute("INSERT INTO system.repairs "
-                            "(parent_id, cfids, coordinator, last_update, participants, ranges, repaired_at, started_at, state) "
-                            "VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)",
-                            [session_id, {cfid}, node1.address(), now, {n.address() for n in self.cluster.nodelist()},
+                            "(parent_id, cfids, coordinator, coordinator_port, last_update, participants, participants_wp, ranges, repaired_at, started_at, state) "
+                            "VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
+                            [session_id, {cfid}, node1.address(), 7000, now, {n.address() for n in self.cluster.nodelist()},
+                             {str(n.address()) + ":7000" for n in self.cluster.nodelist()},
```
jasobrown: is hardcoding a port of `7000` ok here?

requirements.txt
```diff
@@ -4,7 +4,7 @@
 futures
 six
 -e git+https://github.com/datastax/python-driver.git@cassandra-test#egg=cassandra-driver
-ccm==3.1.0
+-e git+https://github.com/aweisberg/ccm.git@master#egg=ccm
```
jasobrown: probably don't want to commit this when we're ready :)

replace_address_test.py
```diff
@@ -35,7 +35,7 @@ class BaseReplaceAddressTest(Tester):
 
     def _setup(self, n=3, opts=None, enable_byteman=False, mixed_versions=False):
         debug("Starting cluster with {} nodes.".format(n))
-        self.cluster.populate(n)
+        self.cluster.populate(n, use_vnodes=not DISABLE_VNODES)
```
jasobrown: Is this change necessary? It's not clear to me that vnodes would have an effect on configurable ports

secondary_indexes_test.py
```diff
@@ -1153,6 +1153,7 @@ def __init__(self, *args, **kwargs):
             r'Streaming error occurred',
             r'\[Stream.*\] Streaming error occurred',
             r'\[Stream.*\] Remote peer 127.0.0.\d failed stream session',
+            r'\[Stream.*\] Remote peer 127.0.0.\d:7000 failed stream session',
```
jasobrown: is the hardcoded port OK here?

snitch_test.py
```diff
@@ -35,15 +36,29 @@ def test_prefer_local_reconnect_on_listen_address(self):
         NODE1_LISTEN_ADDRESS = '127.0.0.1'
         NODE1_BROADCAST_ADDRESS = '127.0.0.3'
 
+        NODE1_LISTEN_FMT_ADDRESS = '/127.0.0.1'
+        NODE1_BROADCAST_FMT_ADDRESS = '/127.0.0.3'
+
+        NODE1_40_LISTEN_ADDRESS = '127.0.0.1:7000'
```
jasobrown: are the hard coded ports ok here?

topology_test.py
```diff
@@ -197,7 +197,7 @@ def resumable_decommission_test(self):
 
         Test decommission operation is resumable
         """
-        self.ignore_log_patterns = [r'Streaming error occurred', r'Error while decommissioning node', r'Remote peer 127.0.0.2 failed stream session']
+        self.ignore_log_patterns = [r'Streaming error occurred', r'Error while decommissioning node', r'Remote peer 127.0.0.2 failed stream session', r'Remote peer 127.0.0.2:7000 failed stream session']
```
jasobrown: hard coded port?

jasobrown (commented):
I'm pretty good with this patch. I wonder if it's ok to have some hard coded IPs and ports in the code (primarily when grepping logs for errors related to specific peers), or if we need to get the node's proper addr from the Node object and plug that into the grep search string.

replace_address_test.py
aweisberg: I think I was fixing some other annoying bug. Should have added a comment. I'll test without it and report back.

aboudreault:
@jasobrown Do you know if there was any dtests merged for CASSANDRA-7544?

aweisberg:
My recollection is that for this you would run all the dtests with a single network interface and that would be the test. It was blocked waiting for the Python driver to work with multiple instances on a single interface.

I think there is an environment variable or something that would make ccm automatically switch to a single interface or some other changes to the dtests that let you run them that way. Don't recall.

aboudreault:
@aweisberg That's what I thought... I will see if I can find some time to finish this work and get this merged. Thanks!!

aweisberg:
Not sure about this PR, but my recollection is that ccm and dtests were both able to run and pass with a single network interface. All the tests were updated to check for IP + port correctly.

I think once the Python driver that the dtests use can work with a single interface the only thing to do is run them with a single interface and see if any tests were added/changes in a way that only works with multiple interfaces.

aboudreault:
ah, I see that all tests are indeed already updated in master. The Python driver cassandra-test branch will be updated soon and I will start from there to retest everything. Feel free to close this one then.

After that, I will see what I need to add for https://issues.apache.org/jira/browse/CASSANDRA-15677.

PR #17, opened 2018-02-02
Cassandra 14206 - fix loose version and some of the remaining broken dtests

aweisberg:


spodkowinski:
Do we still need to merge this to fix existing dtest issues? Tested some of the tests locally and they seem to work already. @aweisberg @beobal

PR #27, opened 2018-05-13
Use pycodestyle-2.3.1 and check dependency compatibility with `pip check`

michaelsembwever:
It seems that pycodestyle-2.4.0 is incompatible with flake8


ptnapoleon:
+1

PR #35, opened 2018-08-15
14408

aweisberg:


byteman/stop_reads.btm
```diff
@@ -0,0 +1,8 @@
+# block mutation verb
+RULE disable mutations
+CLASS org.apache.cassandra.db.ReadCommandVerbHandler
+METHOD doVerb
```
aweisberg: Test comment to see if it's mirrored to the work log

aweisberg (commented):
This seems to be missing tests for incremental repair. I also think we aren't necessarily testing all of the permutations of different code paths that were added in this patch. Like what about a six node cluster where each range isn't replicated (either transiently or fully) by every single node?

When I do a second pass and look at the unit tests I'll keep in mind what the dtests should look like.

It seems also like data wise we should insert enough data to populate each range. For my tests I used the OPP with text strings and manually placed tokens so I could conceptually understand what should be streamed where after each repair. I shut down some nodes so not every token from each range was present.

PR #37, opened 2018-09-14
Add tests for CASSANDRA-14145

beobal:
* Enable read time repaired data tracking for all tests by default
* Add some specific tests for mismatching/matching repaired digests

### DIFF
--- a/repair_tests/incremental_repair_test.py
+++ b/repair_tests/incremental_repair_test.py
@@ -18,6 +18,7 @@ from dtest import Tester, create_ks, create_cf
 from tools.assertions import assert_almost_equal, assert_one
 from tools.data import insert_c1c2
 from tools.misc import new_node, ImmutableMapping
+from tools.jmxutils import make_mbean, JolokiaAgent, remove_perf_disable_shared_mem
 
 since = pytest.mark.since
 logger = logging.getLogger(__name__)
@@ -207,6 +208,7 @@ class TestIncRepair(Tester):
         self.fixture_dtest_setup.setup_overrides.cluster_options = ImmutableMapping({'hinted_handoff_enabled': 'false',
                                                                                      'num_tokens': 1,
                                                                                      'commitlog_sync_period_in_ms': 500})
+        self.init_default_config()
         self.cluster.populate(3).start()
         node1, node2, node3 = self.cluster.nodelist()
 
@@ -246,6 +248,7 @@ class TestIncRepair(Tester):
                                                                                      'num_tokens': 1,
                                                                                      'commitlog_sync_period_in_ms': 500})
 
+        self.init_default_config()
         self.cluster.populate(3).start()
         node1, node2, node3 = self.cluster.nodelist()
 
@@ -289,6 +292,7 @@ class TestIncRepair(Tester):
         self.fixture_dtest_setup.setup_overrides.cluster_options = ImmutableMapping({'hinted_handoff_enabled': 'false',
                                                                                      'num_tokens': 1,
                                                                                      'commitlog_sync_period_in_ms': 500})
+        self.init_default_config()
         self.cluster.populate(3).start()
         node1, node2, node3 = self.cluster.nodelist()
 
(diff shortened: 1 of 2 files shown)


repair_tests/incremental_repair_test.py
```diff
@@ -918,3 +931,196 @@ def test_subrange(self):
...
+        if expect_confirmed_inconsistencies:
+            assert cc_after > cc_before
+        else:
+            assert cc_after == cc_before
+
+    def get_attribute_count(self, jmx, bean):
+        # the MBean may not have been initialized, in which case Jolokia agent will return
+        # a HTTP 404 response. If we receive such, we know that the count can only be 0
+        if jmx.has_mbean(bean):
+            # expect 0 digest mismatches
```
jrwest: should this comment be in the else branch? Or removed?

repair_tests/incremental_repair_test.py
```diff
@@ -918,3 +931,196 @@ def test_subrange(self):
...
+        with JolokiaAgent(node1) as jmx:
+            self.query_and_check_repaired_mismatches(jmx, session, "SELECT * FROM ks.tbl WHERE k = 5")
+            self.query_and_check_repaired_mismatches(jmx, session, "SELECT * FROM ks.tbl WHERE k = 5 AND c = 5")
+            # no digest reads for range queries so read repair metric isn't incremented
+            self.query_and_check_repaired_mismatches(jmx, session, "SELECT * FROM ks.tbl", expect_read_repair=False)
+
+    @since('4.0')
+    def test_repaired_tracking_with_mismatching_replicas(self):
+        """
+        there are two variants of this, for single partition slice & names reads and range reads
```
jrwest: This test would benefit from an in-depth description like the ones you added to the other tests.

beobal: Sorry, that was an oversight, added a description.

repair_tests/incremental_repair_test.py
beobal: Removed, it was a hangover from an earlier version of the test where only the digest mismatch count was read via JMX

repair_tests/incremental_repair_test.py
```diff
@@ -918,3 +931,196 @@ def test_subrange(self):
...
+        stmt = SimpleStatement(query)
+        stmt.consistency_level = ConsistencyLevel.ALL
+        session.execute(stmt)
+
+        rr_after = self.get_attribute_count(jmx, rr_count)
+        uc_after = self.get_attribute_count(jmx, unconfirmed_count)
+        cc_after = self.get_attribute_count(jmx, confirmed_count)
+
+        logger.debug("RR: {before}, {after}".format(before=rr_before, after=rr_after))
+        logger.debug("UI: {before}, {after}".format(before=uc_before, after=uc_after))
```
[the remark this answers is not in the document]
beobal: Done

repair_tests/incremental_repair_test.py
```diff
@@ -207,6 +208,7 @@ def test_manual_session_fail(self):
         self.fixture_dtest_setup.setup_overrides.cluster_options = ImmutableMapping({'hinted_handoff_enabled': 'false',
                                                                                      'num_tokens': 1,
                                                                                      'commitlog_sync_period_in_ms': 500})
+        self.fixture_dtest_setup.init_default_config()
```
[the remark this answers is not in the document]
beobal: > I believe `self.init_default_config()` and `self.fixture_dtest_setup.init_default_config()` are synonymous

Hmm yeah, so it is. I find that a little non-obvious and easy to miss, but ¯\_(ツ)_/¯

> Is it because the config changes in the preceding lines weren't actually being picked up prior?

Exactly, the fixture initialization order means that the cluster config has already been set by this point and so none of these options were being used in the existing tests.

beobal:
Thanks @jrwest, pushed a followup commit and replied to most comments inline.

> Two of the new tests are failing for me.

Yep, some of the refactoring in CASSANDRA-14698 broke repaired tracking where the read repair strategy for the table is `NONE`, which is the case in these tests. CASSANDRA-14755 has a fix for this and is patch available.

>  Is there a build with this branch showing how the dtests fare w/ tracking enabled for all of them?

The dtest runs on CASSANDRA-14145 were all done with this branch, so they have tracking enabled (also the CI runs linked from CASSANDRA-14755).

repair_tests/incremental_repair_test.py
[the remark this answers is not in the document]
jrwest: Yea the proxying to the fixture is a bit non-intuitive. iirc it was added to make the transition from nosetest -> pytest easier.

beobal:
CASSANDRA-14755 is committed now, so I've rebased and squashed this & will commit manually.

PR #39, opened 2018-10-05
Add no-read assert to read-repair test

ifesdjeen:


read_repair_test.py
```diff
@@ -495,6 +495,9 @@ def test_normal_read_repair(self):
             assert storage_proxy.speculated_rr_read == 0
             assert storage_proxy.speculated_rr_write == 0
 
+        warn = node2.grep_log("Message-to-self TYPE:READ")
```
aweisberg: Is this something we want to check for in all dtests all the time as part of checking the log for errors?

It might help us avoid leaving spammy log lines that don't mean much to users in the long run.

ifesdjeen: Ultimately - yes. But unfortunately there are still a couple of places we still do that. Maybe we should widen the scope of the ticket and check more places where this is happening. Or open a separate issue.

ifesdjeen:
Thank you for the review, committed the latest version manually

PR #42, opened 2019-02-05
Fix token_generator_test for Cassandra 2.2

jolynch:
In python 3 subprocess returns bytes, so just decode them.

Needed for https://issues.apache.org/jira/browse/CASSANDRA-15012


jolynch:
Merged in 4df308dfe7bbc5fcf5b649fc1e4a3d6bf6daa7ee

PR #43, opened 2019-02-12
Fix assertion for range query

beobal:
See CASSANDRA-15019
The assertion which verifies that differences in between replicas' repaired data sets is incorrect in the range query case. It should be expecting there to be unconfirmed differences, but currently isn't.


beobal:
Merged in 9dc0052

PR #54, opened 2019-10-08
Add cqlsh test for protocol negotiation

beobal:
Also, disable automatic protocol version capping for batch_test

Patch by Sam Tunnicliffe; reviewed by Alex Petrov for CASSANDRA-15193


ifesdjeen:
Thank you for the patch, +1.

PR #59, opened 2020-03-25
Add tests for CASSANDRA-8272 and CASSANDRA-8273

adelapena:


replica_side_filtering_test.py
```diff
@@ -495,4 +509,26 @@ class TestAllowFiltering(ReplicaSideFiltering):
...
+
+    @since('4.0')
+    def test_update_missed_by_transient_replica(self):
+        self._prepare_cluster(
+            create_table="CREATE TABLE t (k int PRIMARY KEY, v text)",
+            both_nodes=["INSERT INTO t(k, v) VALUES (0, 'old')"],
+            only_node1=["UPDATE t SET v = 'new' WHERE k = 0"],
+            transient=True)
+
+        self._assert_none("SELECT * FROM t WHERE v = 'old'")
```
maedhroz, line 522: @adelapena Just to make sure we're on the same page, at `ALL` it is this first query that would fail without RFP, whereas the second query would be fine, correct?

adelapena: That's correct. The second assertion verifies the case where the RFP extra query returns the old value, which is discarded in the second merge round.

PR #62, opened 2020-04-07
Fix cqlsh/copy tests on older branches

stef1927:
This fixes the cqlsh copy tests on branches older than trunk ([CASSANDRA-14050](https://issues.apache.org/jira/browse/CASSANDRA-14050)).

The dependency of the dtests to cqlshlib formatting.py has been removed by running `SELECT *` queries with cqlsh rather than the driver. This avoids applying the formatting manually. In some cases, where formatting was trivial and the number of rows was significant, the driver is still used and the formatting is done manually (normally a conversion to a string was sufficient). This was done to overcome cqlsh paging. Although we can disable paging, parsing a large number of rows could cause memory or speed problems.

### DIFF
--- a/cqlsh_tests/cqlsh_test_types.py
+++ b/cqlsh_tests/cqlsh_test_types.py
@@ -1,29 +1,12 @@
 import datetime
-import sys
+import logging
+import re
 
 from collections import namedtuple
-from contextlib import contextmanager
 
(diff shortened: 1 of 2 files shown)


stef1927:
Committed manually as da8abe3cab3fc186a6cfb2e3771f647a0dac120e

PR #69, opened 2020-05-14
CASSANDRA-13606 Added some log checks on idxs management

bereng:


secondary_indexes_test.py
```diff
@@ -383,6 +388,7 @@ def test_failing_manual_rebuild_index(self):
         assert before_files == after_files
         assert_none(session, """SELECT * FROM system."IndexInfo" WHERE table_name='k'""")
         assert_one(session, "SELECT * FROM k.t WHERE v = 1", [0, 1])
+        assert 2 == len(node.grep_log('became not-writable'))
```
adelapena, line 391: As before, it would be useful to verify that it has indeed become not-writable:
```suggestion
        assert 2 == len(node.grep_log('became not-writable'))
        
        # Insert a row, that row shouldn't be visible by the index because it's not writable
        session.execute("INSERT INTO k.t(k, v) VALUES (4, 1)")
        assert_one(session, "SELECT * FROM k.t WHERE v = 1", [0, 1])
```
This write should be visible as soon as we successfully rebuild the index, as we do a few lines later.

secondary_indexes_test.py
```diff
@@ -394,6 +400,7 @@ def test_failing_manual_rebuild_index(self):
         assert before_files != after_files
         assert_one(session, """SELECT table_name, index_name FROM system."IndexInfo" WHERE table_name='k'""", ['k', 'idx'])
         assert_one(session, "SELECT * FROM k.t WHERE v = 1", [0, 1])
```
adelapena, line 402: If we have done writes while the index wasn't writable, as I suggested above, those writes should be visible here, after the index recovery:
```suggestion
        assert_all(session, "SELECT * FROM k.t WHERE v = 1", [[0, 1], [4, 1], [3, 1]])
```
Unfortunately, the row inserted after the node restart ([4, 1]) will remain not-visible, leaving the index inconsistent with its base table even after a manual recovery. This is because that row was written before the index writable, and it isn't on the sstables yet. I think that, in case of recovery, we need to flush the base table right after setting the index as writable, and before we start indexing the sstables.

This is a problem that apparently we don't see after a failed index creation because index creation is a schema change that produces a flush. However, I wonder if there is a chance to lose data between that schema flush and the registering of the index as writable. If so, that would be independent of the changes introduced by this ticket, and it might be addresses in a separate ticket if we want to limit the scope.

secondary_indexes_test.py
```diff
@@ -357,7 +359,8 @@ def test_failing_manual_rebuild_index(self):
         assert before_files == after_files
         assert_none(session, """SELECT * FROM system."IndexInfo" WHERE table_name='k'""")
         assert_one(session, "SELECT * FROM k.t WHERE v = 1", [0, 1])
-
+        assert 1 == len(node.grep_log('became not-writable'))
```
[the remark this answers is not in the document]
bereng: I will answer in the other PR

secondary_indexes_test.py
bereng: I added a flush on the C* PR

PR #72, opened 2020-06-04
Introduce --execute-upgrade-tests-only to only run upgrade tests

nastra:


conftest.py
```diff
@@ -466,6 +468,11 @@ def pytest_collection_modifyitems(items, config):
     for item in items:
         deselect_test = False
 
+        if config.getoption("--execute-upgrade-tests-only"):
+            deselect_test = True
```
snazy, line 472:
```suggestion
            deselect_test = not item.get_closest_marker("upgrade_test")
```

nastra: fixed

PR #77, opened 2020-06-16
Add tests for CASSANDRA-15459

adelapena:


consistency_test.py
```diff
@@ -1571,6 +1571,78 @@ def test_quorum_available_during_failure(self):
...
+            query += " WITH READ_REPAIR='NONE'"
+        session.execute(query)
+
+        # with node2 down, populate data on node1
+        # node1:
+        #   key 1 : 1
+        #   key 0 : x
+        #   key 2 : 2
+
+        node2.stop()
```
maedhroz, line 1609: @adelapena This is the sort of thing where we might be able to get real performance advantages in an in-JVM dtest, where we can just call `executeInternal()` rather than stopping a node.

adelapena: Right, I have added an in-JVM dtest. We can take back this PR.

adelapena:
Closing without merging because the tests have been done in-JVM

PR #83, opened 2020-07-08
CASSANDRA-15912 Wait on notifications to propagate

bereng:


bereng:
Tested locally both the failure and the fix against 3.0, 3.11 and 4.0

PR #84, opened 2020-07-15
CASSANDRA-15948 Multiple failures in cqlsh_tests.test_cqlsh.TestCqlsh

maedhroz:
changed expectations around whether collection sub-fields of a UDT will show up with the "frozen" modifier in versions that don't support non-frozen UDTs


maedhroz:
Committed as https://github.com/apache/cassandra-dtest/commit/a70e855b27d7e95dd2ca33245422b05c9c1949b6

PR #89, opened 2020-08-12
Add test for the upgrade of KEYS 2i to 4.0 (CASSANDRA-15906)

pcmanus:
Somehow, I had forgotten to commit this when committing [CASSANDRA-15906](https://issues.apache.org/jira/browse/CASSANDRA-15906) so here it is.


pcmanus:
Committed as f25832a.

PR #95, opened 2020-09-23
CASSANDRA 15977 4.0 quality testing: Read Repair

adelapena:
Disable read repair tests included in 4.x in-JVM dtests

### DIFF
--- a/read_repair_test.py
+++ b/read_repair_test.py
@@ -19,6 +19,7 @@ from tools.jmxutils import JolokiaAgent, make_mbean
 from tools.misc import retry_till_success
 
 since = pytest.mark.since
+ported_to_in_jvm = pytest.mark.ported_to_in_jvm
 logger = logging.getLogger(__name__)
 
 def byteman_validate(node, script, verbose=False, opts=None):
@@ -100,6 +101,7 @@ class TestReadRepair(Tester):
         cluster.start()
 
     @since('3.0')
+    @ported_to_in_jvm('4.0')
     def test_alter_rf_and_run_read_repair(self, fixture_set_cluster_settings):
         """
         @jira_ticket CASSANDRA-10655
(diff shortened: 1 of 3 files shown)


maedhroz (approved):

read_repair_test.py
```diff
@@ -732,106 +730,6 @@ def stop_reads(*nodes, kind='all'):
     with _byteman_cycle(nodes, {'data': [data], 'digest': [digest], 'all': [data, digest]}[kind]):
         yield
 
-kcvv = lambda k, c, v1, v2: [k, c, v1, v2]
-
-
-class TestReadRepairGuarantees(Tester):
```
maedhroz, line 738: @adelapena Just to confirm, this is removed entirely because it was added in 4.0 anyway?

adelapena: @maedhroz Yes, we initially did a literal migration of this class that we discarded because we considered that the tested cases were already covered by the existing tests. However, just removing this class feels a bit scary to me, perhaps we could keep it around by now with a deprecation note, WDYT?

maedhroz: I don't see any reason to worry about removing it. If we did all this work over again (transient replication, etc.), we'd probably end up with exactly what we've done in this issue. On the other hand, it doesn't hurt to keep it around for 4.0. I'll defer to your and our second reviewer.

adelapena: @maedhroz as mentioned [on Slack|https://the-asf.slack.com/archives/CK23JSY2K/p1602201135179400], I have created [here](https://github.com/apache/cassandra-dtest/pull/95/commits/1d5c6460a9b5b744c29eacf1f7fddd2dde5ec4d0) a specific Pytest marker to skip tests ported to in-JVM only if we are not using vnodes.

PR #109, opened 2020-12-17
CASSANDRA-16332 - Change upgrade cql 2i test to accept static results coming back

aholmberg:
CASSANDRA-16332

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -4109,6 +4109,7 @@ class TestCQL(UpgradeTester):
 
     def test_static_columns_with_2i(self):
         cursor = self.prepare()
+        initial_version = self.cluster.version()
 
         cursor.execute("""
             CREATE TABLE test (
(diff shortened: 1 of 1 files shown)


upgrade_tests/cql_tests.py
```diff
@@ -4122,6 +4122,7 @@ def test_static_columns_with_2i(self):
 
         cursor.execute("CREATE INDEX ON test(v)")
 
+        from_ver = float(self.UPGRADE_PATH.starting_meta.family)
```
dcapwell: can't family contain stuff like `x`?  should we use LooseVersion like we do in other tests?

checking...

dcapwell: just checked, the following are current family versions

```
family='2.1'
family='3.0'
family='3.11'
family='trunk'
```

so the last one would break as trunk isn't a valid float.  

What we can do is the following

after `self.prepare` do

```
initial_version = self.cluster.version()
```

then change the if check to

```
if initial_version  > LooseVersion('3.0'):
```

aholmberg: I can try that out. I didn't know any upgrade tests start with trunk. Is that a thing?

PR #112, opened 2021-01-15
CASSANDRA-16181 - 4.0 Quality: Replication Test Audit

maedhroz:
add annotations to stop running batch write and hinted handoff tests that have been ported to the in-JVM framework

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -468,52 +468,11 @@ def _skip_ported_msg(current_running_version, ported_from_version):
 @pytest.fixture(autouse=True)
 def fixture_ported_to_in_jvm(request, fixture_dtest_setup):
     """
-    Adds a new mark called 'ported_to_in_jvm' which denotes that a test was ported to jvm-dtest.
+    Adds a new mark called 'ported_to_in_jvm' which denotes that a test was ported to an in-JVM dtest.
 
-    As of this point in time there are weaknesses of jvm-dtest which require these tests to still
-    be run in the cases not covered by jvm-dtest; namely vnode.
+    In-JVM dtests do not currently support running with vnodes, so tests that use this annotation will
+    still be run around those configurations.
     """
-    marker = request.node.get_closest_marker('ported_to_in_jvm')
-    if marker and not request.config.getoption("--use-vnodes"):
-
-        if not marker.args:
-            pytest.skip("ported to in-jvm")
-
-        from_str = marker.args[0]
-        ported_from_version = LooseVersion(from_str)
-
-        # For upgrade tests don't run the test if any of the involved versions
-        # are excluded by the annotation
-        if hasattr(request.cls, "UPGRADE_PATH"):
-            upgrade_path = request.cls.UPGRADE_PATH
-            ccm_repo_cache_dir, _ = ccmlib.repository.setup(upgrade_path.starting_meta.version)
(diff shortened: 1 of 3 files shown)


hintedhandoff_test.py
```diff
@@ -71,6 +71,7 @@ def _do_hinted_handoff(self, node1, node2, enabled, keyspace='ks'):
             else:
                 query_c1c2(session, n, ConsistencyLevel.ONE, tolerate_missing=True, must_be_missing=True)
 
+    @since('3.0', max_version='3.11.x')
```
adelapena: Shouldn't this use `@ported_to_in_jvm`? If we consider that vnodes are not relevant for the test we can also add `@no_vnodes`, this would also skip those tests in 3.0 and 3.11. Anyway, I think that `@no_vnodes` is generally used to skip tests that don't support vnodes, not that don't require them, so runs with and without vnodes are complete.

adelapena (approved):

adelapena (approved):
Changes look good to me, +1

ekaterinadimitrova2 (approved):
LGTM

conftest.py
```diff
@@ -486,47 +486,12 @@ def fixture_ported_to_in_jvm(request, fixture_dtest_setup):
...
                 pytest.skip(skip_msg)
 
+
 def _skip_ported_msg(current_running_version, ported_from_version):
     if loose_version_compare(current_running_version, ported_from_version) >= 0:
         return "ported to in-JVM from %s >= %s" % (ported_from_version, current_running_version)
 
 
-@pytest.fixture(autouse=True)
-def fixture_ported_to_in_jvm(request, fixture_dtest_setup):
```
maedhroz, line 516: @adelapena @ekaterinadimitrova2 This was a character-for-character duplicate of the one above (modulo the comment).

conftest.py
```diff
@@ -448,10 +449,10 @@ def _skip_ported_msg(current_running_version, ported_from_version):
 @pytest.fixture(autouse=True)
 def fixture_ported_to_in_jvm(request, fixture_dtest_setup):
     """
-    Adds a new mark called 'ported_to_in_jvm' which denotes that a test was ported to jvm-dtest.
+    Adds a new mark called 'ported_to_in_jvm' which denotes that a test was ported to an in-JVM dtest.
 
-    As of this point in time there are weaknesses of jvm-dtest which require these tests to still
-    be run in the cases not covered by jvm-dtest; namely vnode.
+    In-JVM dtests do not currently support running with compression or vnodes, so tests that use this annotation will
```
adelapena: It seems that Python dtests aren't run with different compression options anymore, so we can remove the mention of compression, especially since the fixture code doesn't do anything about it.

adelapena: I think that JVM dtests are also missing the Python option `--use-off-heap-memtables`. Perhaps we should add it to this fixture so ported dtests are still run if that option is set, although we could do that in a separate ticket.

maedhroz: I'm fine with the separate ticket option. In the meantime, I'll rebase and remove the mention of compression.

adelapena (approved):

maedhroz:
Committed in https://github.com/apache/cassandra-dtest/commit/f930ecaed698913827e8d32a0003099a1e11a2c1

PR #113, opened 2021-01-19
CASSANDRA-16395. Explicit node start timeouts.

tlasica:
Some tests require longer start timeout than default 90s:
* bootstrap with reset state
* node replacement
* cdc tests (due to checks for other seeds connectivity)

Before: use default timeout, 90s or rather 600s (due to bug in ccm)
After: use explicit timeout per test case: 120s or 180s

### DIFF
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -547,7 +547,7 @@ class TestCDC(Tester):
         logger.debug('adding node')
         self.cluster.add(loading_node, is_seed=True)
         logger.debug('starting new node')
-        loading_node.start(wait_for_binary_proto=True)
+        loading_node.start(wait_for_binary_proto=120)
         logger.debug('recreating ks and table')
         loading_session = self.patient_exclusive_cql_connection(loading_node)
         create_ks(loading_session, ks_name, rf=1)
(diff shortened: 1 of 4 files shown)


cdc_test.py
```diff
@@ -547,7 +547,7 @@ def _init_new_loading_node(self, ks_name, create_stmt, use_thrift=False):
         logger.debug('adding node')
         self.cluster.add(loading_node, is_seed=True)
         logger.debug('starting new node')
-        loading_node.start(wait_for_binary_proto=True)
+        loading_node.start(wait_for_binary_proto=120)
```
tlasica, line 550: instead of increasing timeout we could use some magical flag 
to skip seed node2 seed connection checking (as node1 is down and node2 is also a seed).
this is new "safety guard" in C* 4.0, but I have not found any switch to turn it off...

I was considering to make timeout conditional for 4.0 but this would imo bring a little value with added code complexity

michaelsembwever (approved):

michaelsembwever:
Committed with https://github.com/apache/cassandra-dtest/commit/e8c2b94b6f106e276800aa3de2628a73a70ac5e6

PR #115, opened 2021-01-22
CASSANDRA-16399: Fix collecting resource intensive tests

jacek-lewandowski:
This patch fixes behaviour for both `run_dtests.py` and `pytest`.

In particular:
- Error handling for invalid parameter values / combinations is in a single place (`dtest_config.py`) and is executed before we actually traverse through the tests
- We exit with just a clean error message instead of tons of spam
- `run_dtests.sh` will not loose the exit code of `pytest` any more so we can clearly detect when test cases collection fails
- removed a bit of boilerplate code from `run_dtests.py` - in particular what it did with xml processing is simply provided with `-q` argument of `pytest`
- tests filtering has been refactored to be cleaner
- fixed filtering of resource intensive tests

Note that now `run_dtests.py` seems to be redundant. If we need it only for listing dtests, we can simply achieve exactly the same effect using `--collect-only -q --ignore=meta_tests` arguments for `pytest` instead of `--dtest-print-tests-only`, plus  we need to filter output with `grep '.py::'` (in order to not include the summary line) and pipe stdout to the target file. Anyway I simplified `run_dtests.sh` so that it just use `pytest` with those arguments.

Regarding Cassandra CI, I suppose most of the configurations will benefit with this fix:
 - `no-vnodes`, `vnodes`, `off-heap-memtables` - previously included resource intensive tests although it was specified to skip them in the build script
 - `large` - instead of listing just resource intensive tests it listed all tests

Some examples of error handling now vs before:
- before:
```
$ ./run_dtests.py --dtest-print-tests-only
<?xml version="1.0" encoding="UTF-8"?>
<Modules>
      </Instance>
    </Class>
  </Module>
</Modules>

$ pytest --collect-only
==================================================== test session starts =====================================================
platform darwin -- Python 3.9.1, pytest-3.6.4, py-1.10.0, pluggy-0.7.1
rootdir: /Users/jlewandowski/dev/datastax/cassandra-dtest/trunk, inifile: pytest.ini
plugins: timeout-1.4.2, flaky-3.7.0
timeout: 900.0s
timeout method: signal
timeout func_only: False
collecting 0 items / 1 errors                                                                                                INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/jlewandowski/dev/datastax/cassandra-dtest/trunk/venv/lib/python3.9/site-packages/_pytest/main.py", line 178, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
...
INTERNALERROR> FileNotFoundError: [Errno 2] No such file or directory: 'build.xml'

================================================== 1 error in 0.18 seconds ===================================================

$ ./run_dtests.py --cassandra-dir=../../cassandra/ds-trunk --dtest-print-tests-only --only-resource-intensive-tests --skip-resource-intensive-tests
<?xml version="1.0" encoding="UTF-8"?>
<Modules>
  <Module name="auditlog_test.py">
...
auth_test.py::TestAuthUnavailable::test_permission_cache_background_reload_handle_unavailable
auth_test.py::TestNetworkAuth::test_full_dc_access
auth_test.py::TestNetworkAuth::test_single_dc_access
...
(no error)
```

- now:
```
$ ./run_dtests.py --dtest-print-tests-only
ERROR: Required dtest arguments were missing! You must provide either --cassandra-dir or --cassandra-version. You can also set 'cassandra_dir' in pytest.ini. Refer to the documentation or invoke the help with --help.

$ pytest --collect-only
ERROR: Required dtest arguments were missing! You must provide either --cassandra-dir or --cassandra-version. You can also set 'cassandra_dir' in pytest.ini. Refer to the documentation or invoke the help with --help.

$ ./run_dtests.py --cassandra-dir=../../cassandra/ds-trunk --dtest-print-tests-only --only-resource-intensive-tests --skip-resource-intensive-tests
ERROR: --skip-resource-intensive-tests does not make any sense with either --only-resource-intensive-tests or --force-resource-intensive-tests.

$ pytest --cassandra-dir=../../cassandra/ds-trunk --collect-only --only-resource-intensive-tests --skip-resource-intensive-tests
ERROR: --skip-resource-intensive-tests does not make any sense with either --only-resource-intensive-tests or --force-resource-intensive-tests.
```


conftest.py
```diff
@@ -602,20 +609,17 @@ def pytest_collection_modifyitems(items, config):
...
 
-        if item.get_closest_marker("resource_intensive") and not collect_only:
-            force_resource_intensive = config.getoption("--force-resource-intensive-tests")
-            skip_resource_intensive = config.getoption("--skip-resource-intensive-tests")
-            if not force_resource_intensive:
-                if skip_resource_intensive:
-                    deselect_test = True
-                    logger.info("SKIP: Deselecting test %s as test marked resource_intensive. To force execution of "
-                          "this test re-run with the --force-resource-intensive-tests command line argument" % item.name)
+        if item.get_closest_marker("resource_intensive"):
```
tlasica: this code is so complex that IMO it would be reasonable to add tests in `meta_tests` with acceptance criteria validation, so that one does not have to reverse engineer the rules.

decision making for single `item` should be extracted to either a class or a function so that it can be easily tested without whole configuration burden.

not blocking the fix.

conftest.py
```diff
@@ -594,6 +594,13 @@ def pytest_collection_modifyitems(items, config):
     sufficient_system_resources_resource_intensive = sufficient_system_resources_for_resource_intensive_tests()
     logger.debug("has sufficient resources? %s" % sufficient_system_resources_resource_intensive)
 
+    force_resource_intensive = config.getoption("--force-resource-intensive-tests")
+    skip_resource_intensive = config.getoption("--skip-resource-intensive-tests")
+    only_resource_intensive = config.getoption("--only-resource-intensive-tests")
+    if skip_resource_intensive and only_resource_intensive:
```
tlasica: also `skip` and `force` should not be combined?

conftest.py
```diff
@@ -602,20 +609,17 @@ def pytest_collection_modifyitems(items, config):
...
+                logger.info("SKIP: Deselecting test %s as test marked resource_intensive. To force execution of "
+                            "this test re-run with the --force-resource-intensive-tests command line argument" % item.name)
+            elif not force_resource_intensive:
                 if not sufficient_system_resources_resource_intensive:
                     deselect_test = True
                     logger.info("SKIP: Deselecting resource_intensive test %s due to insufficient system resources" % item.name)
 
-        if not item.get_closest_marker("resource_intensive") and not collect_only:
-            only_resource_intensive = config.getoption("--only-resource-intensive-tests")
+        if not item.get_closest_marker("resource_intensive"):
```
tlasica: I wonder what was the rationale for running this logic only if `not collect_only`?
maybe there was a valid reason not to do it always?
@michaelsembwever ?

tlasica (approved):
with comments

conftest.py
jacek-lewandowski: @tlasica It was already mentioned as a problem in https://issues.apache.org/jira/browse/CASSANDRA-16221

michaelsembwever: Not that I can remember, unfortunately :( 
(though @jacek-lewandowski is correct, the fault has already been illustrated)

michaelsembwever (approved):

michaelsembwever:
Up to https://github.com/apache/cassandra-dtest/pull/115/commits/b4fc4f07c2cc4c2779250f672bc09d9403647bb1 LGTM! 

Given the size of the patch, even if it's simple stuff in practice, it would be nice to see some tests for the different combinations added to `meta_tests/`

conftest.py
```diff
@@ -597,58 +588,59 @@ def pytest_collection_modifyitems(items, config):
...
+        is_upgrade_test = _is_upgrade_test(item)
+        is_resource_intensive_test = item.get_closest_marker("resource_intensive") is not None
+        is_vnodes_test = item.get_closest_marker("vnodes") is not None
+        is_no_vnodes_test = item.get_closest_marker("no_vnodes") is not None
+        is_no_offheap_memtable_test = item.get_closest_marker("no_offheap_memtables") is not None
+        is_depends_driver_test = item.get_closest_marker("depends_driver") is not None
+
+        if is_upgrade_test:
+            if not dtest_config.execute_upgrade_tests and not dtest_config.execute_upgrade_tests_only:
+                deselect_test = True
```
tlasica: nit: I do not see why we keep `deselect_test` and `deseleted` instead of `skip_test` and `skipped"... of course also in messaging.

conftest.py
```diff
@@ -597,58 +588,59 @@ def pytest_collection_modifyitems(items, config):
     for item in items:
         deselect_test = False
 
-        if config.getoption("--execute-upgrade-tests-only"):
-            deselect_test = not item.get_closest_marker("upgrade_test")
-            if deselect_test:
+        is_upgrade_test = _is_upgrade_test(item)
```
tlasica: nit: I think it would not be bad to actually extract:
```
def should_skip_test(test, dtest_config):
```
because one could put `return` statements there which would  not cause reader to reflect if the `deselect` decision can be reverted.. making reasoning easier.


and then the loop will become:
```
for item in items:
   skip_test = self.should_skip_test(item, dtest_config)
   if skip_test:
      ...
   else:
      ....
```

tlasica: or even better:
```
skipped = [x for x in items if self.should_skip_test(x, dtest_config)
selected = list((set(items) - set(skipped))
```
no ifs, no loops, voila.

conftest.py
```diff
@@ -597,58 +588,59 @@ def pytest_collection_modifyitems(items, config):
...
+                if not dtest_config.force_execution_of_resource_intensive_tests:
                     deselect_test = True
-                    logger.info("SKIP: Deselecting resource_intensive test %s due to insufficient system resources" % item.name)
-
-        if not item.get_closest_marker("resource_intensive") and not collect_only:
-            only_resource_intensive = config.getoption("--only-resource-intensive-tests")
-            if only_resource_intensive:
+                    logger.info("SKIP: Deselecting resource_intensive test %s due to insufficient system resources. "
+                                "Specify --force-resource-intensive-tests if you really want to run it" % item.name)
+        else:
```
tlasica: nit: I am not a big fun of nested ifs and elseses...
maybe instead it should be a sequence of rules.
if rules i matched -> skip.
even if repeating conditions.

```
if dtest_config.only_resource_intensive and not is_resource_intentive:
   skip
if dtest_config.use_vnodes and is_no_vnodes_test:
   skip with log
if not dtest_config.use_vnodes and is_vnodes_test:
   skip with log
(...)
```

conftest.py
```diff
@@ -660,5 +652,17 @@ def pytest_collection_modifyitems(items, config):
...
 
-def _upgrade_testing_enabled(config):
-    return config.getoption("--execute-upgrade-tests") or config.getoption("--execute-upgrade-tests-only")
+def _is_upgrade_test(item):
+    if item.get_closest_marker("upgrade_test") is not None:
+        return True
+    else:
+        is_upgrade_test = False
+
+        for test_item_class in inspect.getmembers(item.module, inspect.isclass):
```
tlasica: what will it be? a tuple? then it can be unboxed.
`test_item_class[1]` is rather cryptic ;-)

conftest.py
```diff
@@ -660,5 +652,17 @@ def pytest_collection_modifyitems(items, config):
...
+            if not hasattr(test_item_class[1], "pytestmark"):
+                continue
+            for module_pytest_mark in test_item_class[1].pytestmark:
+                if module_pytest_mark.name == "upgrade_test":
+                    is_upgrade_test = True
```
tlasica: if we want to keep this `for/if` then we can `break` once found.

but we can also go with more like
```
pytest_mark_names = [m.name for m in test_item_class[1].pytestmark
if "upgrade_test" in pytest_mark_names:
    return True
```

tlasica (approved):
with some code improvement comments
up to you if you want to apply

meta_tests/conftest_test.py
```diff
@@ -0,0 +1,87 @@
...
+from unittest import TestCase
+
+from dtest_config import DTestConfig
+from mock import Mock
+from pytest import UsageError
+
+from conftest import *
+
+
+def _response(input, responses, default_response):
```
tlasica: why not `responses.get(input, default_response)` ?
because of logging?

I think we do not need this function at all.

.travis.yml
```diff
@@ -1,21 +1,23 @@
...
   # feed changed lines with no context around them to pycodestyle
   # I know we don't enforce line length but if you introduce
   # 200-char lines you are doing something terribly wrong.
   # lint all files for everything but line length errors
-  - git diff apache/master...HEAD -U0 | pycodestyle --ignore=E501 --diff
+  - git diff apache/trunk...HEAD -U0 | pycodestyle --ignore=E501 --diff
   # lint all files except json_test.py for line length errors
-  - git diff apache/master...HEAD -U0 | pycodestyle --diff --exclude='json_test.py' --exclude='meta_tests/assertion_test.py' --max-line-length=200
+  - git diff apache/trunk...HEAD -U0 | pycodestyle --diff --exclude='json_test.py' --exclude='meta_tests/assertion_test.py' --max-line-length=200
+  - pytest --metatests
```
michaelsembwever, line 22: 💪

.travis.yml
```diff
@@ -1,21 +1,23 @@
...
-  - pip install pycodestyle==2.3.1 flake8
+  - pip install pycodestyle==2.6.0 flake8
   - pip check
+  - CASS_DRIVER_NO_CYTHON=1 pip install -r requirements.txt
 script:
   # we want pyflakes to check all files for unused imports only
   # we use flake8 because it allows us to ignore other warnings
   # exclude the thrift directories - they contain auto-generated code
-  - flake8 --ignore=E501,F811,F812,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .
+  # - flake8 --ignore=E501,F811,F812,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .
```
michaelsembwever, line 12: instead of commenting it out, could we add more ignores (for those currently failing?) 


(duplicate and comment out the line so we know what it is supposed to be)

michaelsembwever (approved):

.travis.yml
jacek-lewandowski: I don't know if you tried, but there are so many of them that is probably make no sense at all - instead, I can spend 1-2 hours on fixing that properly

michaelsembwever: > I can spend 1-2 hours on fixing that properly

i have a PR somewhere for that :-)
(it would need to be a separate ticket.)

let's just comment out the line then for now 👍

meta_tests/dtest_config_test.py
```diff
@@ -0,0 +1,121 @@
...
+    config = Mock()
+    config.getoption.side_effect = _mock_responses(params)
+    config.getini.side_effect = _mock_responses({})
+
+    dTestConfig = DTestConfig()
+    dTestConfig.setup(config)
+    return dTestConfig
+
+
+def _check_with_params_expect(params, pattern):
```
tlasica, line 27: reading the name I would say we always expect exception.
but the code as it is does not check if exception is actually thrown,
it only checks if "when exception is thrown it has proper message".

maybe we should instead use:
```with pytest.raises(ValueError, match=r".* 123 .*"):```

tlasica (approved):
Great job!

michaelsembwever:
Merged with https://github.com/apache/cassandra-dtest/commit/ec84618b7450ef9357a3a88fc93e39d74a34b02e

PR #117, opened 2021-01-26
CASSANDRA-16296 Node can join when RF > N in multi-DC

bereng:


bootstrap_test.py
```diff
@@ -280,6 +280,28 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+        node5 = Node('node5',
+                     cluster,
+                     True,
+                     ('127.0.0.5', 9160),
+                     ('127.0.0.5', 7000),
+                     '7500',
+                     '0',
+                     None,
+                     binary_interface=('127.0.0.5', 9042))
+        cluster.add(node5, is_seed=False, data_center="dc1")
```
bereng: Once c16411 merges we should remove "dc1" here as a test for 16411 #justfyi

bootstrap_test.py
```diff
@@ -280,6 +280,28 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+        cluster = self.cluster
+        cluster.set_environment_variable('CASSANDRA_TOKEN_PREGENERATION_DISABLED', 'True')
+        cluster.populate([2, 2])
+        cluster.start()
+
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+        session.execute("CREATE KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '3'}")
+
+        node5 = Node('node5',
```
tlasica: nit: using parameter names would make it probably easier to read

bootstrap_test.py
tlasica: Probably too much, but
should we add a failed read before / successful read after addition?

bootstrap_test.py
```diff
@@ -280,6 +280,28 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
     def test_consistent_range_movement_false_with_rf1_should_succeed(self):
         self._bootstrap_test_with_replica_down(False, rf=1)
 
+    def test_RF_GT_Nodes_should_succeed(self):
```
tlasica: is having multiple DCs essential for this problem?
If so then maybe it should be reflected in the test name?

tlasica (approved):
with some minor questions

bootstrap_test.py
bereng: Yep multi-dc is the key. I can change the name.

bootstrap_test.py
bereng: It won't hurt...

bootstrap_test.py
```diff
@@ -280,6 +280,33 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+        cluster = self.cluster
+        cluster.set_environment_variable('CASSANDRA_TOKEN_PREGENERATION_DISABLED', 'True')
+        cluster.populate([1, 1])
+        cluster.start()
+
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+        session.execute("CREATE KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '2'}")
+        session.execute("CREATE TABLE k.testgtrfmultidc (KEY text PRIMARY KEY)")
+        session.execute("INSERT INTO k.testgtrfmultidc (KEY) VALUES ('test_rf_gt_nodes_multidc_should_succeed')")
```
adelapena, line 305: We could easily verify here that the log warnings are emitted in both nodes, since we don't check that in unit tests:
```python
warning = 'Your replication factor 2 for keyspace k is higher than the number of nodes 1 for datacenter dc1'
if cluster.version() >= '4.0':
    for node in cluster.nodelist():
        assert len(node.grep_log(warning)) == 3
```
Then we could verify that the warning is not emitted for the third node, since we have reached a cluster size that matches the RF:
```python
if cluster.version() >= '4.0':
    assert len(node3.grep_log(warning)) == 0
```

bootstrap_test.py
```diff
@@ -280,6 +280,33 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
     def test_consistent_range_movement_false_with_rf1_should_succeed(self):
         self._bootstrap_test_with_replica_down(False, rf=1)
 
+    def test_rf_gt_nodes_multidc_should_succeed(self):
```
adelapena, line 283: We could add some brief docstring mentioning the purpose of the test and mentioning the related ticket(s) with `@jira_ticket`.

bootstrap_test.py
bereng: +1

bootstrap_test.py
```diff
@@ -280,6 +280,44 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
     def test_consistent_range_movement_false_with_rf1_should_succeed(self):
         self._bootstrap_test_with_replica_down(False, rf=1)
 
+    """
+    Test creating a KS with RF > N on multi DC doesn't break bootstrap
+    @jira_ticket CASSANDRA-16296 CASSANDRA16411
+    """
+    def test_rf_gt_nodes_multidc_should_succeed(self):
```
adelapena: Nit: I think doctring should be the first statement in the function's body. There is also a missed hyphen in the second ticket id:
```suggestion
    def test_rf_gt_nodes_multidc_should_succeed(self):
        """
        Test creating a KS with RF > N on multi DC doesn't break bootstrap
        @jira_ticket CASSANDRA-16296 CASSANDRA-16411
        """
```

bootstrap_test.py
```diff
@@ -280,6 +280,53 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+        cluster = self.cluster
+        cluster.set_environment_variable('CASSANDRA_TOKEN_PREGENERATION_DISABLED', 'True')
+        cluster.populate([1, 1])
+        cluster.start()
+
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+        session.execute("CREATE KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '3'}")
+        warning = 'Your replication factor 3 for keyspace k is higher than the number of nodes 1 for datacenter dc1'
+        warnings_total = 0
```
adelapena: Nit: this could be into the `if` block body. The same applies for the second initialization of `warnings_total` below.

bootstrap_test.py
```diff
@@ -280,6 +280,53 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
     def test_consistent_range_movement_false_with_rf1_should_succeed(self):
         self._bootstrap_test_with_replica_down(False, rf=1)
 
+    def test_rf_gt_nodes_multidc_should_succeed(self):
+        """
+        Test creating a KS with RF > N on multi DC doesn't break bootstrap
```
adelapena: Nit: now it's creating and altering a keyspace.

adelapena (approved):
The new test looks good to me, I have just left a couple of trivial suggestions.

bootstrap_test.py
bereng: The gist imo is about KS creation. The `Alter` bit is just operational commands you do to it after creation. It's relevant to the warnings logic but not to the main problem. Would you agree?

bootstrap_test.py
```diff
@@ -280,6 +280,55 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+        cluster.populate([1, 1])
+        cluster.start()
+
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+        session.execute("CREATE KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '3'}")
+        warning = 'Your replication factor 3 for keyspace k is higher than the number of nodes 1 for datacenter dc1'
+
+        if cluster.version() >= '4.0':
+            warnings_total = 0
```
tlasica:
```
warnings_total = sum([len(node.grep_log(warning) for node in cluster.node_list()])
assert warnings_total == 1
```
but actually what we would like to check is: "warning on a dc1 node, and no warning on dc2 node"
so maybe more explicit:
```
assert len(node1.grep_log(warning)) == 1
assert len(node2.grep_log(warning)) == 0
```

bootstrap_test.py
```diff
@@ -280,6 +280,55 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+                warnings_total += len(node.grep_log(warning))
+            assert warnings_total == 1
+
+        session.execute("ALTER KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '2'}")
+        session.execute("CREATE TABLE k.testgtrfmultidc (KEY text PRIMARY KEY)")
+        session.execute("INSERT INTO k.testgtrfmultidc (KEY) VALUES ('test_rf_gt_nodes_multidc_should_succeed')")
+        warning = 'Your replication factor 2 for keyspace k is higher than the number of nodes 1 for datacenter dc1'
+
+        if cluster.version() >= '4.0':
+            warnings_total = 0
```
tlasica: same here

bootstrap_test.py
```diff
@@ -280,6 +280,55 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+                     thrift_interface=('127.0.0.3', 9160),
+                     storage_interface=('127.0.0.3', 7000),
+                     jmx_port='7300',
+                     remote_debug_port='0',
+                     initial_token=None,
+                     binary_interface=('127.0.0.3', 9042))
+        cluster.add(node3, is_seed=False, data_center="dc1")
+        node3.start(wait_for_binary_proto=True);
+        if cluster.version() >= '4.0':
+            assert len(node3.grep_log(warning)) == 0
```
tlasica: this check for warning can be missed because number of nodes changed.
so for example `of nodes 1` will not be found if the warning would be `of nodes 2` (as we added the node).

bootstrap_test.py
tlasica: so maybe rather then describing what test does we should describe what it validates?
"Test validating that a KS with RF > N on multi DC doesn not break a bootsrtap"

tlasica (changes requested):

bootstrap_test.py
adelapena: Good point, that makes more sense. Just note that the session created with `patient_cql_connection` can use any node as coordinator so doing the asserts that way would be flaky unless we also create the session with `patient_exclusive_cql_connection`. We can also put `warning` inside the `if` block since I think we are not going to reuse it:
```python
node1, node2 = cluster.nodelist()
session = self.patient_exclusive_cql_connection(node1)
session.execute("CREATE KEYSPACE k WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : '3'}")

if cluster.version() >= '4.0':
    warning = 'Your replication factor 3 for keyspace k is higher than the number of nodes 1 for datacenter dc1'
    assert len(node1.grep_log(warning)) == 1
    assert len(node2.grep_log(warning)) == 0
```

bootstrap_test.py
adelapena: We could just mark the logs before adding the node and search in every node for a generic `'Your replication factor'` message, for example something like:
```python
marks = map(lambda n: n.mark_log(), cluster.nodelist())

# add node...

if cluster.version() >= '4.0':
    warning = 'Your replication factor'
    for (node, mark) in zip(cluster.nodelist(), marks):
        assert len(node.grep_log(warning, from_mark=mark)) == 0
```

bootstrap_test.py
adelapena: "validating that a KS with RF > N..." sounds good to me.

bootstrap_test.py
bereng: Oh I didn't know about the 'exclusive' cql connection gtk.

bootstrap_test.py
bereng: Amended but I used a more particular substring for the check to avoid false matches.

bootstrap_test.py
```diff
@@ -280,6 +280,55 @@ def test_consistent_range_movement_true_with_rf1_should_fail(self):
...
+                     remote_debug_port='0',
+                     initial_token=None,
+                     binary_interface=('127.0.0.3', 9042))
+        cluster.add(node3, is_seed=False, data_center="dc1")
+        node3.start(wait_for_binary_proto=True)
+        if cluster.version() >= '4.0':
+            warning = 'is higher than the number of nodes'
+            for (node, mark) in zip(cluster.nodelist(), marks):
+                assert len(node.grep_log(warning, from_mark=mark)) == 0
+
```
ekaterinadimitrova2, line 328: While testing, I observed that node3 gets flooded with a lot of those(not part of this patch but made me wonder why so many....):
```
WARN  [MigrationStage:1] 2021-02-10 23:32:53,370 MigrationCoordinator.java:429 - Can't send schema pull request: node /127.0.0.1:7000 is down.
WARN  [MigrationStage:1] 2021-02-10 23:32:53,371 MigrationCoordinator.java:429 - Can't send schema pull request: node /127.0.0.2:7000 is down.
```
/cc @adelapena @tlasica

bereng: Maybe we could use the no spam logger?

ekaterinadimitrova2: I was also wondering about that

ekaterinadimitrova2 (approved):

tlasica (approved):

michaelsembwever:
Merged with https://github.com/apache/cassandra-dtest/commit/493ddae492a9a9a47bc484a7dfa75ef86fd3d9b8

PR #120, opened 2021-01-27
Cassandra 16405

tlasica:
This change is preparing for ccm changes, when `NodeError` can be thrown by `ccm` instead of `TimeoutError` when the node does not start due to problem such as wrong config, crc etc.

In addition it fixes some code insights `flake8` issues.

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -11,7 +11,7 @@ import logging
 from cassandra import WriteTimeout
 from cassandra.cluster import NoHostAvailable, OperationTimedOut
 from ccmlib.common import is_win
-from ccmlib.node import Node, TimeoutError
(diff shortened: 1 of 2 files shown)


michaelsembwever:
Thank you for keeping the flake8 commit separate. Keep it in the PR, but i will commit it separately as it's not related to the ticket number.

michaelsembwever (approved):

michaelsembwever:
Merged with https://github.com/apache/cassandra-dtest/commit/2a721e7b058025f42f8b86c8a2ecd851c2790f00 and https://github.com/apache/cassandra-dtest/commit/ebb59dd7cf0c4fb4a9d9135cf6442e59c109de31

PR #131, opened 2021-03-22
CASSANDRA-16530 fix flaky test_ttl_deletions

bereng:


paging_test.py
```diff
@@ -3189,7 +3189,7 @@ def check_all_paging_results(self, expected_data, pagecount, num_page_results):
                                range(0, len(expected_data), page_size)]
 
         pf = self.get_page_fetcher()
-        pf.request_all()
+        pf.request_all(timeout=60)
```
ekaterinadimitrova2, line 3192: Why did you decide to change this one?

bereng: The default timeout is 5s. In a super slow env I managed to make it fail here but only once out of many runs. The original TTL was 3s which is too close to 5s etc. So better put some distance between all timeouts as well.

paging_test.py
```diff
@@ -3406,15 +3406,15 @@ def test_ttl_deletions(self):
...
+                 "values ({}, '{}', {}, {}, {}) using ttl 15;").format(
                 row['id'], row['mytext'], row['col1'],
                 row['col2'], row['col3'])
             self.session.execute(
                 SimpleStatement(s, consistency_level=CL.ALL)
             )
         self.check_all_paging_results(data, 8,
                                       [25, 25, 25, 25, 25, 25, 25, 25])
-        time.sleep(5)
+        time.sleep(15)
```
adelapena: Besides incrementing the time, I like the idea of waiting exactly the TTL time. Perhaps we could make that more obvious by putting the TTL in a var, to be used both in the queries and in the sleep:
```python
ttl_seconds = 15
for row in data:
    s = ("insert into paging_test (id, mytext, col1, col2, col3) "
         "values ({}, '{}', {}, {}, {}) using ttl {}").format(
        row['id'], row['mytext'], row['col1'], row['col2'], row['col3'], ttl_seconds)
    self.session.execute(SimpleStatement(s, consistency_level=CL.ALL))
self.check_all_paging_results(data, 8, [25, 25, 25, 25, 25, 25, 25, 25])
time.sleep(ttl_seconds)
self.check_all_paging_results([], 0, [])
```

adelapena (approved):

paging_test.py
ekaterinadimitrova2: I don't see this one being default 5s? What do I miss? The other one is clear

    ```
def request_all(self, timeout=None):
        """
        Requests any remaining pages.

        If the future is exhausted, this is a no-op.
        @param timeout Time, in seconds, to wait for all pages.    def request_all(self, timeout=None):
        """
        Requests any remaining pages.

        If the future is exhausted, this is a no-op.
        @param timeout Time, in seconds, to wait for all pages.
```

adelapena: @ekaterinadimitrova2 I think the 5s default is [here](https://github.com/apache/cassandra-dtest/blob/trunk/tools/paging.py#L106), although it's not very clear nor documented, and possibly it would be preferable to use Python defaults.

ekaterinadimitrova2: Oh I see; thank you for pointing it to me. Yes, it sounds a bit weird; I would probably just change the None to 5s default or use Python defaults as you mentioned.

adelapena: If the new query timeout is 60s, it might happen that the first query requesting the rows before waiting for the TTL gets them after they have expired. So perhaps it would make sense to use a TTL that is greater or equal than the first query timeout, possibly adding a timeout argument to `check_all_paging_results` which value in `test_ttl_deletions` could be the TTL, wdyt?

ekaterinadimitrova2: Sounds reasonable to me

paging_test.py
bereng: +1

paging_test.py
bereng: I'm not following fully. Ping me on Slack when you're online. I don't see a problem in returning a 'snapshot' of how things were _at query time_. If rows TTL mid-way through the query execution I don't care. I care about the time between insert and query being less than TTL.

adelapena: Just ignore me. I was wrong interpreting how pagination is done. I thought that the subqueries for each page in the first query were going to have its own timestamp, so for example a paged query with a timestamp=1 would generate subqueries with timestamp=1, timestamp=1+n, etc. This would lead to paging subqueries starting too late to be sure of getting the rows before the TTL expiration. Fortunately it seems that the pager implementation is much clever than that and it uses the same timestamp for all the subqueries in a paged query, so this is not a problem.

ekaterinadimitrova2: :D Seems I made the same mistake

ekaterinadimitrova2 (approved):

PR #133, opened 2021-04-21
Remove references to run_dtests from README

mfleming:
Newcomers to cassandra-dtest that look through `README.md` will see that the `run_dtests.py` script is the quickest way to get started running tests. Unfortunately, the script has a number of problems and I'm not sure it ever work properly after the move to the pytest framework.

## Process stdout/stderr buffering
Firstly, when I execute `run_dtests.py` I don't see any output after
```
$ ./run_dtests.py --dtest-tests paging_test.py 
============================= test session starts ==============================
```

This looks likely to be because of the buffering that pytest does internally for stdout and stderr and because of the way that it's executed by `run_dtests.py`, i.e. I suspect that `run_dtests.py` is blocked on the following line for most of the execution because there's no data available in the pipe for stderr:
```
stderr_output = sp.stderr.readline()
```

See also https://github.com/pytest-dev/pytest/issues/1886

## `--pytest-options` doesn't work
Secondly, the options specified in `--pytest-options` aren't actually passed through to pytest.

## Most devs run pytest directly
When I spoke to @ekaterinadimitrova2 it seemed like most developers just run the tests directly with pytest which would explain why `run_dtests.py` has bitrotted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,9 @@ path to success as it provides common base setup across various configurations.
 Usage
 -----
 
-The tests are executed by the pytest framework. For convenience, a wrapper ``run_dtests.py`` 
-is included with the intent to make starting execution of the dtests with sane defaults as easy 
-as possible. Most users will most likely find that invoking the tests directly using ``pytest`` 
-ultimately works the best and provides the most flexibility.
-
-Pytest has a great [Usage and Invocations](https://docs.pytest.org/en/latest/usage.html) document which is a great place to start for basic invocation options when using pytest.
+The tests are executed by the pytest framework which includes a helpful [Usage and
+Invocations](https://docs.pytest.org/en/latest/usage.html) document which is a great place to start
+for basic invocation options when using pytest.
 
 At minimum, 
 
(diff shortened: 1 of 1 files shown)


michaelsembwever:
It is used for listing tests still. For example see [here](https://github.com/apache/cassandra-builds/blob/trunk/build-scripts/cassandra-dtest-pytest.sh#L89)

But the changes to the README in this PR LGTM.

mfleming:
> I was thinking a bit about this. I don't think we should remove it but improve the wording.
> Probably we can say explicitly that developers should use pytest as this is just to run the tests getting back minimal info. WDYT?

It doesn't return any info from my testing other than the "test session starts" line, e.g.
```
$ ./run_dtests.py --dtest-tests cql_prepared_test.py 
============================= test session starts ==============================
$ 
```
To me (as a newbie) this doesn't seem like a particularly user-friendly way to run the tests at all. It doesn't tell me where the logs live, doesn't provide any info on the test progress or even if the test passed/failed. In fact, if I patch `cql_prepared_test.py` to make the test fail immediately, I see the exact same output as above.

ekaterinadimitrova2 (approved):

PR #135, opened 2021-05-10
Flaky TestTransientReplicationRing.test_move_backwards_and_cleanup

bereng:


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # git tag -a -f cassandra-test
 # git push origin :refs/tags/cassandra-test
 # git push -f origin refs/tags/cassandra-test
--e git+https://github.com/riptano/ccm.git@cassandra-test#egg=ccm
+-e git+https://github.com/ekaterinadimitrova2/ccm.git@16644#egg=ccm
```
ekaterinadimitrova2: Return the riptano CCM, I was using my repo only for test purposes.

transient_replication_ring_test.py
```diff
@@ -226,13 +227,12 @@ def test_bootstrap_and_cleanup(self):
         # Every value should be replicated exactly 2 times
         self.check_replication(sessions, exactly=2)
 
-    @flaky(max_runs=1)
```
ekaterinadimitrova2, line 229: Yes, I was talking about this one but please, remove it for all tests in the class. It is the same useless and don't seem right to me to leave it there. Thanks

requirements.txt
bereng: This I already mentioned in the ticket

PR #147, opened 2021-06-22
16733 followup

blerer:


ekaterinadimitrova2 (approved):
So my understanding is that now we test for the default behavior which is DROP COMPACT STORAGE disabled, LGTM +1

PR #149, opened 2021-07-05
Update timer metrics assertion for Cassandra-16760

yifan-c:


client_request_metrics_test.py
```diff
@@ -777,6 +783,11 @@ def is_histo_list(k, l):
     assert len(l) == 165, k
     assert all(isinstance(i, int) for i in l), k
 
+def is_timer_histo_list(k, l):
```
maedhroz: nit: Maybe we could just add a named parameter to `is_histo_list` (ex. `is_histo_list(k, l, len)`) and use `partial()` above at the call site? (see https://docs.python.org/3/library/functools.html#functools.partial)

maedhroz (approved):

PR #152, opened 2021-08-11
CASSANDRA-16841 Fix unexpectedly ignored dtests

k-rus:
This commit fixes few unexpected behaviors, which were:
1. If any test class is marked with `resource_intensive`, then all tests
   in the file were treated as resource intensive.
2. If one test class is marked with `vnodes` and another test class is
   marked with `no_vnodes` in the same file, then all tests were
   skipped in the file.

The unexpected behaviors are due to marking all tests in a file if any 
class is assigned the mark. This commit fixes that marks on other 
classes in the file do not affect a test, which is not marked and 
neither its class. This is also applied to upgrade tests, so the code 
is simplified and behavior is unified.

Since the upgrade tests were utilizing that the marking one class is 
propagated to all classes in the file, they were revisited. In many 
cases the upgrade tests were already included, because their classes 
inherited from a class, which already was marked. In such cases 
unnecessary marks were removed.

The unit tests of `is_shippable` from `conftest.py` are changed to test
calculating skipping conditions from provided mocked configurations
instead of shipping if conditions are satisfied or not, which required
to change the underlying implementation. The unit tests are also
rewritten to be parameterized and it covers now all possible meaningful
test cases.

patch by Ruslan Fomkin; reviewed by Andrés de la Peña and Ekaterina 
Dimitrova for CASSANDRA-16841

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -28,7 +28,8 @@ since = pytest.mark.since
 ported_to_in_jvm = pytest.mark.ported_to_in_jvm
 logger = logging.getLogger(__name__)
 
-class TestBootstrap(Tester):
+
+class BootstrapTester(Tester):
     byteman_submit_path_pre_4_0 = './byteman/pre4.0/stream_failure.btm'
     byteman_submit_path_4_0 = './byteman/4.0/stream_failure.btm'
 
@@ -1097,3 +1098,13 @@ class TestBootstrap(Tester):
         # 3. check host_id in other node's table
         session1 = self.patient_exclusive_cql_connection(node1)
         assert_one(session1, "SELECT host_id FROM system.peers_v2 WHERE peer = {}".format(address2), [uuid.UUID(host_id)])
+
+
+class TestBootstrap(BootstrapTester):
+    """
+    This child class is a helper for PyTest to pick up the test methods.
+    Since the same test methods are executed as part of upgrade, this helper child class is
+    necessary to avoid double execution of the same tests.
+    It is necessary to put some dummy expression in this child class.
+    """
+    pass
--- a/conftest.py
+++ b/conftest.py
@@ -548,9 +548,12 @@ class SkipConditions:
     def __init__(self, dtest_config, sufficient_resources):
         self.skip_upgrade_tests = not dtest_config.execute_upgrade_tests and not dtest_config.execute_upgrade_tests_only
         self.skip_non_upgrade_tests = dtest_config.execute_upgrade_tests_only
-        self.skip_resource_intensive_tests = (
+        self.skip_resource_intensive_due_to_resources = (
             not dtest_config.force_execution_of_resource_intensive_tests
-            and not sufficient_resources) or dtest_config.skip_resource_intensive_tests
+            and not sufficient_resources)
+        self.skip_resource_intensive_tests = (
+            self.skip_resource_intensive_due_to_resources
+            or dtest_config.skip_resource_intensive_tests)
         self.skip_non_resource_intensive_tests = dtest_config.only_resource_intensive_tests
         self.skip_vnodes_tests = not dtest_config.use_vnodes
         self.skip_no_vnodes_tests = dtest_config.use_vnodes
@@ -606,10 +609,10 @@ def pytest_collection_modifyitems(items, config):
     sufficient_resources = sufficient_system_resources_for_resource_intensive_tests()
     skip_conditions = SkipConditions(dtest_config, sufficient_resources)
 
-    if skip_conditions.skip_resource_intensive_tests:
+    if skip_conditions.skip_resource_intensive_due_to_resources:
         logger.info("Resource intensive tests will be skipped because "
-                    "it was requested to skip or there is not enough system resource "
-                    "and --force-resource-intensive-tests were not specified")
+                    "there is not enough system resources "
+                    "and --force-resource-intensive-tests was not specified")
 
     for item in items:
         deselect_test = SkipConditions.is_skippable(skip_conditions, item)
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,5 @@
 import copy
 import collections
-import inspect
 import logging
 import os
 import platform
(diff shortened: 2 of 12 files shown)


k-rus:
[CircleCI build](https://app.circleci.com/pipelines/github/k-rus/cassandra?branch=CASS-16841-CI) is running

conftest.py
```diff
@@ -545,53 +544,46 @@ def cassandra_dir_and_version(config):
...
-
-def _is_skippable(item, mark, include_marked, include_other):
-    if has_mark(item, mark):
-        if include_marked:
-            return False
-        else:
-            logger.info("SKIP: Skipping %s because it is marked with %s" % (item, mark))
+        if skip_marked:
+            logger.info(
+                "SKIP: Skipping %s because it is marked with %s" % (item, mark))
```
adelapena: Nit: I think we don't need to break this line nor the other `loggger.info` call right below
```suggestion
            logger.info("SKIP: Skipping %s because it is marked with %s" % (item, mark))
```

conftest.py
```diff
@@ -545,53 +544,46 @@ def cassandra_dir_and_version(config):
...
-    return skippable
+def is_skippable(item, skip_cond):
+    return (_is_skippable(item, "upgrade_test",
+                          skip_cond.skip_upgrade_tests, skip_cond.skip_non_upgrade_tests)
+            or _is_skippable(item, "resource_intensive",
+                             skip_cond.skip_resource_intensive_tests, skip_cond.skip_non_resource_intensive_tests)
+            or _is_skippable(item, "vnodes", skip_cond.skip_vnodes_tests, skip_non_marked=False)
+            or _is_skippable(item, "no_vnodes", skip_cond.skip_no_vnodes_tests, skip_non_marked=False)
+            or _is_skippable(item, "no_offheap_memtables", skip_cond.skip_no_offheap_memtables_tests, skip_non_marked=False)
+            or _is_skippable(item, "depends_driver", skip_marked=True, skip_non_marked=False))
```
adelapena: Nit: maybe this block would be a bit easier to read if we always name the `skip_marked`/`skip_non_marked` parameters:
```suggestion
    return (self._is_skippable(item, "upgrade_test",
                               skip_marked=self.skip_upgrade_tests,
                               skip_non_marked=self.skip_non_upgrade_tests)
            or self._is_skippable(item, "resource_intensive",
                                  skip_marked=self.skip_resource_intensive_tests,
                                  skip_non_marked=self.skip_non_resource_intensive_tests)
            or self._is_skippable(item, "vnodes",
                                  skip_marked=self.skip_vnodes_tests,
                                  skip_non_marked=False)
            or self._is_skippable(item, "no_vnodes",
                                  skip_marked=self.skip_no_vnodes_tests,
                                  skip_non_marked=False)
            or self._is_skippable(item, "no_offheap_memtables",
                                  skip_marked=self.skip_no_offheap_memtables_tests,
                                  skip_non_marked=False)
            or self._is_skippable(item, "depends_driver",
                                  skip_marked=True,
                                  skip_non_marked=False))
```

conftest.py
```diff
@@ -545,53 +544,46 @@ def cassandra_dir_and_version(config):
...
-                 include_upgrade_tests,
-                 include_non_upgrade_tests,
-                 include_resource_intensive_tests,
-                 include_non_resource_intensive_tests,
-                 include_vnodes_tests,
-                 include_no_vnodes_tests,
-                 include_no_offheap_memtables_tests):
-
-    skippable = False
+class SkipConditions:
```
adelapena, line 547: If we are going to pack these vars into a new class perhaps we could also place `_is_skippable` and `is_skippable` inside the class, for example [this way](https://github.com/adelapena/cassandra-dtest/commit/bb3b4382a59c453ccfcd8b85c5cbc70afe8b5cee).

upgrade_tests/paging_test.py
```diff
@@ -188,6 +188,7 @@ def random_txt(text):
             assert_lists_equal_ignoring_order(pf.all_data(), expected_data, sort_key='value')
 
 
+@pytest.mark.upgrade_test
```
adelapena: Can we simply add the annotation to `BasePagingTester`, which is the parent of all the marked tests, [this way](https://github.com/adelapena/cassandra-dtest/commit/c3668b62654b0b5f11eccb9e2cfd74d1e164d990)?

.gitignore
```diff
@@ -10,3 +10,4 @@ last_test_dir
 upgrade
 html/
 doxygen/doxypy-0.4.2/
+.pytest_cache/
```
adelapena: +1000

conftest.py
```diff
@@ -605,28 +597,17 @@ def pytest_collection_modifyitems(items, config):
...
-    include_resource_intensive_tests = can_run_resource_intensive_tests and not dtest_config.skip_resource_intensive_tests
-    include_non_resource_intensive_tests = not dtest_config.only_resource_intensive_tests
-    include_vnodes_tests = dtest_config.use_vnodes
-    include_no_vnodes_tests = not dtest_config.use_vnodes
-    include_no_offheap_memtables_tests = not dtest_config.use_off_heap_memtables
+    if skip_conditions.skip_resource_intensive_tests:
+        logger.info("Resource intensive tests will be skipped because "
+                    "it was requested to skip or there is not enough system resource "
+                    "and --force-resource-intensive-tests or --only-resource-intensive-tests "
+                    "were not specified")
```
adelapena: I don't see why skipping resource intensive tests with not enough resources and `--only-resource-intensive-tests` is wrong. IMO it makes sense to require the `--force-resource-intensive-tests` flag in that case, since one flag controls the selection and the other one skips the resource check. Let's see what the second reviewer thinks about this. In any case if we are going to change the behaviour of the flags or we find them confusing we should probably extend their descriptions in `pytest_addoption` and `run_dtests.py`.

conftest.py
k-rus: It is not required. I guess I had default setting for autopep8. I will restore it back.

conftest.py
k-rus: I explained my thoughts in the Jira ticket, so I guess we continue there and let's see what others think about it.

upgrade_tests/paging_test.py
k-rus: Thank you for pointing it. Actually marking `BasePagingTester` is not needed, since it inherits form `UpgradeTester`, which is already marked. My main concern is that the marking becomes implicit. I don't know if it will be confusing or not, while  believe developers will have intention that all tests are marked as upgrade tests in this folder. I can try to clarify in `upgrade_tests/README.md` that using `UpgradeTester` is enough.

I tested that it works in `pytest`, however the readme instructions use `nosetests` to run upgrade tests.

conftest.py
k-rus: Thank you for the suggestion with the patch. Looks nice. I plan to incorporate it into the patch.

k-rus:
I addressed comments and started [new build in CircleCI](https://app.circleci.com/pipelines/github/k-rus/cassandra?branch=CASS-16841-CI).

upgrade_tests/paging_test.py
adelapena: I think that the note in the readme makes this clearer, thanks. At some point we will have to clear the references to nosetests, but I think there is no need to do it during this ticket.

k-rus:
Addressed comment and run [j11_dtest build](https://app.circleci.com/pipelines/github/k-rus/cassandra/10/workflows/e56bdedd-45ba-466f-ae32-a003b9004063)

conftest.py
```diff
@@ -605,28 +603,16 @@ def pytest_collection_modifyitems(items, config):
...
-    include_non_upgrade_tests = not dtest_config.execute_upgrade_tests_only
-    include_resource_intensive_tests = can_run_resource_intensive_tests and not dtest_config.skip_resource_intensive_tests
-    include_non_resource_intensive_tests = not dtest_config.only_resource_intensive_tests
-    include_vnodes_tests = dtest_config.use_vnodes
-    include_no_vnodes_tests = not dtest_config.use_vnodes
-    include_no_offheap_memtables_tests = not dtest_config.use_off_heap_memtables
+    if skip_conditions.skip_resource_intensive_tests:
+        logger.info("Resource intensive tests will be skipped because "
+                    "it was requested to skip or there is not enough system resource "
+                    "and --force-resource-intensive-tests were not specified")
```
adelapena:
```suggestion
                    "and --force-resource-intensive-tests was not specified")
```

k-rus: Thank you for catching this!

bootstrap_test.py
```diff
@@ -1097,3 +1097,12 @@ def test_host_id_override(self):
         # 3. check host_id in other node's table
         session1 = self.patient_exclusive_cql_connection(node1)
         assert_one(session1, "SELECT host_id FROM system.peers_v2 WHERE peer = {}".format(address2), [uuid.UUID(host_id)])
+
+class TestBootstrap(BootstrapTester):
```
adelapena, line 1103: Nit: missed additional blank line before the class declaration

k-rus: Thank you for noting! I haven't set up automatic formatting on commits and there is nothing prepared in Cassandra to just reuse. I know that there is git commit hook, but it doesn't fix formatting, it is just erroring.

And took almost a day to get Travis to run it.

I will fix it soon.

ekaterinadimitrova2: I use PyCharm and I get warnings for formatting, never had to do anything in addition to be able to see those.

k-rus: I use VS Code. My preferable approach is to be not dependent on IDE, thus I am looking into setting up a git commit hook.

PR #160, opened 2021-09-14
Node reusage

bereng:


conftest.py
```diff
@@ -122,6 +127,35 @@ def fixture_dtest_setup_overrides(dtest_config):
...
+
+@pytest.fixture(scope='session', autouse=True)
+def fixture_dtest_clean_reused_cluster(request):
+    """
+    Clean re-used cluster at end of session
+    """
+
+    yield
+
+    global reusable_dtest_setup
```
jmckenzie-dev, line 157: Please comment why the global at at least one of the usage sites of it, or preferably refactor out the need / usage of this.

conftest.py
```diff
@@ -322,20 +356,60 @@ def fixture_dtest_setup(request,
                         fixture_dtest_setup_overrides,
                         fixture_logging_setup,
                         fixture_dtest_cluster_name,
-                        fixture_dtest_create_cluster_func):
-    if running_in_docker():
-        cleanup_docker_environment_before_test_execution()
+                        fixture_dtest_create_cluster_func,
+                        fixture_dtest_reuse_cluster):
+    #TODO Reuse nodes taking into account
```
jmckenzie-dev, line 361: Status on TODO?

conftest.py
```diff
@@ -367,7 +447,87 @@ def fixture_dtest_setup(request,
...
             logger.error("Error saving log:", str(e))
         finally:
+            if failed or reuse_option == Reuse_cluster.REUSE_CLUSTER_NO:
+                safe_cluster_cleanup(request, dtest_setup)
+                reusable_dtest_setup = None
+
+
+def cant_reuse_cluster_reason(request, dtest_setup, last_test_class, dtest_config):
+    if dtest_setup is None:
+        return "Can't reuse bc reusable cluster is None"
```
jmckenzie-dev, line 457: Could use a more descriptive error message - what does it mean for dtest_setup to be None for example?

conftest.py
```diff
@@ -367,7 +447,87 @@ def fixture_dtest_setup(request,
...
+
+    # When passing tests as line items to pytest a 'class' fixture won't work hence this hack
+    current_test_class = str(request.cls.__name__)
+    if last_test_class is not None and last_test_class != current_test_class:
+        return "Can't reuse bc new test class: " + last_test_class + " -> " + current_test_class
+
+    #Only reuse nodes starting at 4.0
+    current_vesion = dtest_config.cassandra_version_from_build
+    if loose_version_compare(LooseVersion('4.0.1'), current_vesion) > 0:
+        return "Can't reuse bc current version: " + str(current_vesion) + " is not gt 4.0.1 "
```
jmckenzie-dev, line 473: Should this be "is not >= 4.0"?

conftest.py
```diff
@@ -122,6 +127,35 @@ def fixture_dtest_setup_overrides(dtest_config):
...
+        return Reuse_cluster.REUSE_CLUSTER_YES
+
+
+@pytest.fixture(scope='session', autouse=True)
+def fixture_dtest_clean_reused_cluster(request):
+    """
+    Clean re-used cluster at end of session
+    """
+
+    yield
```
jmckenzie-dev, line 155: Document the expected flow that'll come in here and why the yield is appropriate. This looks to be pretty subtle and pytest specific (referencing: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended), so either class level doc on how we're using this or at least locally annotating the what / why flow would be a useful breadcrumb for subsequent maintainers.

conftest.py
```diff
@@ -322,20 +356,60 @@ def fixture_dtest_setup(request,
...
-    if running_in_docker():
-        cleanup_docker_environment_before_test_execution()
+                        fixture_dtest_create_cluster_func,
+                        fixture_dtest_reuse_cluster):
+    #TODO Reuse nodes taking into account
+    #Byteman cleanup
+    #Auto detect configuration options
+
+    reuse_option = fixture_dtest_reuse_cluster
+    # Use reuse_option = Reuse_cluster.REUSE_CLUSTER_NO to fully disable node-reusage
```
jmckenzie-dev, line 366: Can we parse env vars instead of having .py modification required to configure this feature? We should default to this feature being disabled and run a bespoke job on ASF CI for a bit to iron out the kinks rather than default in.

conftest.py
```diff
@@ -322,20 +356,60 @@ def fixture_dtest_setup(request,
                         fixture_dtest_setup_overrides,
                         fixture_logging_setup,
                         fixture_dtest_cluster_name,
-                        fixture_dtest_create_cluster_func):
-    if running_in_docker():
-        cleanup_docker_environment_before_test_execution()
+                        fixture_dtest_create_cluster_func,
+                        fixture_dtest_reuse_cluster):
+    #TODO Reuse nodes taking into account
+    #Byteman cleanup
```
jmckenzie-dev, line 362: What're these two lines here for?

conftest.py
```diff
@@ -367,7 +447,87 @@ def fixture_dtest_setup(request,
...
+        if len(dtest_setup.cluster.nodelist()) != 0:
+            node1 = dtest_setup.cluster.nodelist()[0]
+            if node1.is_running():
+                # Clear KS
+                session = dtest_setup.cql_connection(node1, user='cassandra', password='cassandra')
+                query = "select * from system_schema.keyspaces"
+                rows = session.execute(query=query, timeout=120)
+                for row in rows:
+                    ks = str(row.keyspace_name)
+                    if not ks.startswith("system"):
```
jmckenzie-dev, line 500: Thinking through this, would it be possible or desired to snapshot the state of the system keyspace upon node creation and revert it to that clean state after tests? I assume a node would get _quite unhappy_ were we to go in mucking around w/a nuke and pave of system for instance, but perhaps some logic to snapshot the state of a table and revert it.

May be a lot harder than I'm thinking, and if this works fine as-is for the tests you've set it up for I suppose we're ok? Just seems like the contents of the system keyspace pre and post this patch are going to differ which worries me.

conftest.py
```diff
@@ -367,7 +447,87 @@ def fixture_dtest_setup(request,
...
+
+def setup_cluster(dtest_config,
+                  fixture_dtest_setup_overrides,
+                  fixture_dtest_cluster_name,
+                  fixture_dtest_create_cluster_func,
+                  reuse_cluster = False):
+    if running_in_docker():
+        cleanup_docker_environment_before_test_execution(reuse_cluster)
+
+    # do all of our setup operations to get the enviornment ready for the actual test
```
jmckenzie-dev, line 519: nit: spelling. "environment"

conftest.py
```diff
@@ -359,6 +437,8 @@ def fixture_dtest_setup(request,
                 failed = True
                 pytest.fail(msg='Unexpected error found in node logs (see stdout for full details). Errors: [{errors}]'
                             .format(errors=str.join(", ", errors)), pytrace=False)
+        for node in dtest_setup.cluster.nodelist():
```
jmckenzie-dev, line 440: I'd leave a comment informing other maintainers _why_ we're doing this here just to be extra clear (re-using node, only want to have space of N logs visible for error grepping, etc etc)

conftest.py
```diff
@@ -367,7 +447,87 @@ def fixture_dtest_setup(request,
...
+                safe_cluster_cleanup(request, dtest_setup)
+                reusable_dtest_setup = None
+
+
+def cant_reuse_cluster_reason(request, dtest_setup, last_test_class, dtest_config):
+    if dtest_setup is None:
+        return "Can't reuse bc reusable cluster is None"
+
+    # Can't reuse if some nodes down
+    if dtest_setup is not None \
```
jmckenzie-dev, line 460: Recommend rewriting this to something a bit more clear. For example:
```
    # Can't reuse if some nodes down
    live_node_count = len([node for node in dtest_setup.cluster.nodelist() if node.is_live()])
    if dtest_setup is not None and live_node_count != len(dtest_setup.cluster.nodelist()):
        return "Can't reuse cluster because number of live node count doesn't match"```

File: README.md [hunk 12]
```diff
@@ -82,14 +82,28 @@ the ``--count`` option. For example, to run a test class 10 times:
...
 tests.
 
+Node reusage 4.0+
+-----------------
 Each test spawns a new fresh cluster and tears it down after the test. If a
 test fails, the logs for the node are saved in a `logs/<timestamp>` directory
 for analysis (it's not perfect but has been good enough so far, I'm open to
 better suggestions).
 
+Starting at 4.0, as dtests can be very heavy, in order to speed up things node reusage has been implemented. Basically a
```
jmckenzie-dev, line 99: Recommend rephrasing this - the above paragraph states "Each test spawns a new fresh cluster..." and then this paragraph refutes it. Consider something like "In the default configuration, each test spawns a new fresh cluster", then revising this paragraph to read "Optionally starting at 4.0...".

We'd also need to make this patch optional to make that true as well, but I left another comment about how I think that's the right path.

splitter.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -x
```
jmckenzie-dev, line 3: In general, this script could use a header comment and some sugar sprinkled throughout to help guide people on why it's doing what it's doing.

For example: what is a `chunk`? How does that differ from a `split`? etc.

cqlsh_tests/test_cqlsh.py
```diff
@@ -127,6 +128,810 @@ def tearDown(self):
...
+        assert describe_out == reloaded_describe_out
+
+        reloaded_select_out, err = self.run_cqlsh(node1, "SELECT * FROM test.users_by_state")
+        assert 0 == len(err), err
+        assert select_out == reloaded_select_out
+
+
+@reuse_cluster
+class TestCqlshReuse(TestCqlshBase):
+    def try_reuse_node(self):
```
jmckenzie-dev, line 930: Looks like we should unify on a single idiom as we have another pattern in `auth_test.py` with `clear_auth`.

Maybe something like `reset_cluster_between_tests` that becomes the idiomatic implementation we use? Or better yet, is called by default if you have the @reuse_cluster annotation specified? If we did that, we could define a default implementation that raises an exception in Tester so if someone tries to call it without remembering to extend that in their test class they get a hard failure / signal they're not using things correctly.

cqlsh_tests/test_cqlsh.py
```diff
@@ -2587,19 +2575,37 @@ def test_cjk_output(self):
...
     """
+    def clear_auth(self, session):
+        # Clear auth after test to support node reusage
+        rows = list(session.execute(query="list roles", timeout=120))
+        for row in rows:
+            role = str(row.role)
+            if role != 'cassandra':
+                logging.debug("drop role if exists '" + role + "'")
+                session.execute(query="drop role if exists '" + role + "'", timeout=120)
+            session.cluster.control_connection.wait_for_schema_agreement(wait_time=120)
```
jmckenzie-dev, line 2591: Why are we waiting on a schema agreement here when we didn't change schema and just dropped specific roles?

paging_test.py
```diff
@@ -3134,11 +3141,7 @@ def random_txt(text):
         self.assertEqualIgnoreOrder(flatten_into_set(page_fetchers[10].all_data()), flatten_into_set(expected_data[:50000]))
 
 
-@since('2.0')
-class TestPagingWithDeletions(BasePagingTester, PageAssertionMixin):
-    """
-    Tests concerned with paging when deletions occur.
-    """
+class TestPagingWithDeletionsBase(BasePagingTester, PageAssertionMixin):
```
jmckenzie-dev, line 3144: Accidental drop of `@since` annotation?

replica_side_filtering_test.py
```diff
@@ -116,7 +123,10 @@ def create_index(self):
         """
         pass
 
-    def test_update_on_skinny_table(self):
+
+class ReplicaSideFilteringBaseTests(ReplicaSideFilteringBase):
+
+    def testupdate_on_skinny_table(self):
```
jmckenzie-dev, line 129: Accidental rename from test_update?

replica_side_filtering_test.py
```diff
@@ -177,7 +187,7 @@ def test_update_on_collection(self):
         self._assert_none("SELECT * FROM t WHERE v CONTAINS 0")
         self._assert_one("SELECT * FROM t WHERE v CONTAINS 1", row=[0, [-1, 1]])
 
-    def test_complementary_deletion_with_limit_on_partition_key_column_with_empty_partitions(self):
+    def testcomplementary_deletion_with_limit_on_partition_key_column_with_empty_partitions(self):
```
jmckenzie-dev, line 190: Looks like we lost a space here as well with test_complementary

replica_side_filtering_test.py
```diff
@@ -550,9 +573,9 @@ def _test_missed_update_with_transient_replicas(self, missed_by_transient):
         self._assert_one("SELECT * FROM t WHERE v = 'new'", row=[0, 'new'])
 
     @since('4.0')
-    def test_update_missed_by_transient_replica(self):
+    def testupdate_missed_by_transient_replica(self):
```
jmckenzie-dev, line 576: Revert to test_update

user_functions_test.py
```diff
@@ -33,16 +31,24 @@ def fixture_dtest_setup_overrides(self, dtest_config):
...
 
-        cluster.populate(nodes).start()
-        node1 = cluster.nodelist()[0]
-        time.sleep(0.2)
+        if len(cluster.nodelist()) == 0:
+            cluster.populate(nodes).start()
+            node1 = cluster.nodelist()[0]
+            time.sleep(0.2)
+        else:
+            node1 = cluster.nodelist()[0]
```
jmckenzie-dev, line 41: Can drop else and just move node1 assignment out here.

materialized_views_test.py
```diff
@@ -53,12 +54,16 @@ def _rows_to_list(self, rows):
...
+
+        if len(cluster.nodelist()) == 0:
+            cluster.set_configuration_options({'enable_materialized_views': 'true'})
+            cluster.populate([nodes, 0], install_byteman=install_byteman)
+            if options:
+                cluster.set_configuration_options(values=options)
+            cluster.start()
+            node1 = cluster.nodelist()[0]
+        else:
+            node1 = cluster.nodelist()[0]
```
jmckenzie-dev: Some permutation of this keeps showing up in every test. Can we have something in Tester that has a signature like:
`Node[] maybe_populate_cluster(count, config_stuff)` and we use it like:
`node1, node2, node3 = maybe_populate_cluster(count, config_stuff)` so we can get around repeating this boilerplate for every reusable test? Could trigger that before each run if we have the `@reuse_cluster` annotation set so tests wouldn't even really need to have any of this here.

jmckenzie-dev (changes requested):

PR #161, opened 2021-09-16
Fix flaky test: repair_tests/repair_test.py::TestRepair::test_local_d…

ifesdjeen:
…c_repair

Patch by Alex Petrov; reviewed by TBD for CASSANDRA-16963

### DIFF
--- a/tools/data.py
+++ b/tools/data.py
@@ -17,14 +17,17 @@ def create_c1c2_table(tester, session, read_repair=None):
     create_cf(session, 'cf', columns={'c1': 'text', 'c2': 'text'}, read_repair=read_repair)
 
 
-def insert_c1c2(session, keys=None, n=None, consistency=ConsistencyLevel.QUORUM):
+def insert_c1c2(session, ks=None, keys=None, n=None, consistency=ConsistencyLevel.QUORUM):
     if (keys is None and n is None) or (keys is not None and n is not None):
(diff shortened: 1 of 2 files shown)


tools/data.py
```diff
@@ -17,14 +17,17 @@ def create_c1c2_table(tester, session, read_repair=None):
...
+def insert_c1c2(session, ks=None, keys=None, n=None, consistency=ConsistencyLevel.QUORUM):
     if (keys is None and n is None) or (keys is not None and n is not None):
         raise ValueError("Expected exactly one of 'keys' or 'n' arguments to not be None; "
                          "got keys={keys}, n={n}".format(keys=keys, n=n))
     if n:
         keys = list(range(n))
 
-    statement = session.prepare("INSERT INTO cf (key, c1, c2) VALUES (?, 'value1', 'value2')")
+    fully_qualified_cf = "cf"
+    if ((ks is not None) and (not ks)):
```
maedhroz: Doesn't `not ks` mean the string is empty? i.e. This conditional says "ks is not None but it is empty"

Do we want something like `if ((ks is not None) and ks != ""):` or even `if ((ks is not None) and not(not ks)):`?

ifesdjeen: You're right, it should've been a double negation!

maedhroz (approved):

maedhroz (approved):

PR #163, opened 2021-10-04
Updated existing tests and added a new test for CASSANDRA-16983

bowensong:
Please see: [CASSANDRA-16983](https://issues.apache.org/jira/browse/CASSANDRA-16983) and the [pull request #1220](https://github.com/apache/cassandra/pull/1220) for apache/cassandra.

### DIFF
--- a/cqlsh_tests/test_cqlsh.py
+++ b/cqlsh_tests/test_cqlsh.py
@@ -8,6 +8,7 @@ import datetime
 import locale
 import os
 import re
+import stat
 import subprocess
 import sys
 import logging
@@ -43,7 +44,11 @@ class CqlshMixin():
     cqlsh_debug_enabled = False
 
     def verify_output(self, query, node, expected):
-        output, err = self.run_cqlsh(node, query, ['-u', 'cassandra', '-p', 'cassandra'])
+        cqlsh_options = ['-u', 'cassandra', '-p', 'cassandra']
+        if self.cluster.version() >= '4.1':
+            cqlsh_options.append('--insecure-password-without-warning')
+
+        output, err = self.run_cqlsh(node, query, cqlsh_options)
         if common.is_win():
             output = output.replace('\r', '')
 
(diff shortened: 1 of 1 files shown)


cqlsh_tests/test_cqlsh.py
```diff
@@ -2184,6 +2189,102 @@ def test_fetch_all_rows_in_batch_mode(self):
...
+            # Ensure the deprecation warning shows when user and/or password is found in the cqlshrc file
+            cqlshrcfile.seek(0)
+            cqlshrcfile.truncate()
+            cqlshrcfile.write('[authentication]\nusername=cassandra\n')
+            cqlshrcfile.flush()
+            os.chmod(cqlshrcfile.name, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+            cqlsh_options = ['--cqlshrc', cqlshrcfile.name, '-p', 'cassandra', '--insecure-password-without-warning']
+            _, cqlsh_stderr = self.run_cqlsh(node1, cmds='quit', cqlsh_options=cqlsh_options)
+            assert 'insecure cqlshrc file' not in cqlsh_stderr
+            assert 'Credentials in the cqlshrc file is deprecated' in cqlsh_stderr
```
smiklosovic, line 2233: you could please rephrase this?

"Putting your credentials into the cqlshrc file is deprecated. Please move them to dedicated file ..." something like this.

It is good to say not what the problem is but how to actually fix it.

bowensong: The message is in two lines, and the assert only checks for a partial match on the the first line. The full message is:

Notice: Credentials in the cqlshrc file is deprecated and will be ignored in the future.
Please use a credentials file to specify the username and password.

Which I believe is exactly you had expected - not only describe the problem, but also how to fix it.

Here's the link to the changed cqlsh.py in the cassandra repo: https://github.com/apache/cassandra/blob/cd62e34a6133a2f5f8066c93654119114795faba/bin/cqlsh.py#L2153

bhouse99 (approved):
Looks like it answers the behavior we are adding quite well.

Ran the tests for the pr, and after some hiccups, no errors

PR #165, opened 2021-10-09
CASSANDRA-16334 Test write failures in multi-DC setup

Gerrrr:


write_failures_test.py
```diff
@@ -228,3 +229,46 @@ def test_thrift(self):
...
+        )
+
+    def _test_oversized_mutation(self, consistency_level):
+        """
+        Test that multi-DC write failures return operation failed rather than a timeout.
+        @jira_ticket CASSANDRA-16334.
+        """
+
+        cluster = self.cluster
+        cluster.populate([3, 3])
```
adelapena: Do we need 3+3 nodes? I think we could use less nodes here to save some resources

write_failures_test.py
```diff
@@ -228,3 +229,46 @@ def test_thrift(self):
...
+
+        session.execute("CREATE KEYSPACE test WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': 3, 'dc2': 3};")
+        session.execute("CREATE TABLE test.test (key int PRIMARY KEY, val blob);")
+
+        payload = '1' * 1024 * 256
+        statement = SimpleStatement("""
+            INSERT INTO test.test (key, val) VALUES (1, textAsBlob('{}'))
+            """.format(payload), consistency_level=consistency_level)
+
+        with pytest.raises(WriteFailure) as cm:
```
adelapena: Nit: we don't use the `as cm` part.

write_failures_test.py
```diff
@@ -228,3 +229,46 @@ def test_thrift(self):
...
+            session.execute(statement)
+
+    def test_oversized_mutation_local_one(self):
+        self._test_oversized_mutation(ConsistencyLevel.LOCAL_ONE)
+
+    def test_oversized_mutation_one(self):
```
adelapena: Having a test per consistency level is nice and clear, but maybe we could combine both tests to use a single cluster and so save some resources, for example [this way](https://github.com/adelapena/cassandra-dtest/commit/19fb3e719d208fa286fa80f86567ddd99628d18b). This and the reduction of nodes make the test 3-4 times faster, at least locally. wdyt?

write_failures_test.py
```diff
@@ -228,3 +229,46 @@ def test_thrift(self):
...
+        session.execute("CREATE KEYSPACE test WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': 3, 'dc2': 3};")
+        session.execute("CREATE TABLE test.test (key int PRIMARY KEY, val blob);")
+
+        payload = '1' * 1024 * 256
+        statement = SimpleStatement("""
+            INSERT INTO test.test (key, val) VALUES (1, textAsBlob('{}'))
+            """.format(payload), consistency_level=consistency_level)
+
+        with pytest.raises(WriteFailure) as cm:
+            session.execute(statement)
```
adelapena: We could also check that no hints have been created after the failed write, for example using metrics [this way](https://github.com/adelapena/cassandra-dtest/commit/897c2c39b74c73111009ed71204ac9d2c73424f5).

Gerrrr:
I like both suggestions; cherry-picked them into the PR.

PR #169, opened 2021-12-02
Cassandra 15234 take2

ekaterinadimitrova2:


michaelsembwever (approved):
(nit: i'm not a fan of code lines changing solely for the sake of wrapping/format)

PR #185, opened 2022-04-20
Fix test expectations for CASSANDRA-17034

blambov:


maedhroz (approved):

snapshot_test.py
```diff
@@ -270,7 +270,14 @@ def test_archive_commitlog_point_in_time_with_active_commitlog_ln(self):
...
 
-    def run_archive_commitlog(self, restore_point_in_time=False, restore_archived_commitlog=True, archive_active_commitlogs=False, archive_command='cp'):
+    @since('4.1')
+    def test_archive_commitlog_restore_skip_by_position(self):
+        """
+        Test archive commit log not restored because of specified snapshot commit log position
+        """
+        self.run_archive_commitlog(restore_point_in_time=True, archive_active_commitlogs=True, specify_commitlog_position=True, archive_command='ln')
+
+    def run_archive_commitlog(self, restore_point_in_time=False, restore_archived_commitlog=True, archive_active_commitlogs=False, specify_commitlog_position=False, archive_command='cp'):
```
adelapena: Nit: can we split this line to not exceed 120 chars? By the way, it's not related to the patch so we don't have to address it here, but it seems that `archive_active_commitlogs` is not used since CASSANDRA-12509, and the tests `test_archive_commitlog_point_in_time` and `test_archive_commitlog_point_in_time_with_active_commitlog` do the same thing.

adelapena (approved):

PR #192, opened 2022-05-03
CASSANDRA-17594: Fix flaky python-tests due to connection getting closed

dcapwell:


dtest_setup.py
```diff
@@ -72,6 +77,13 @@ def __init__(self, dtest_config=None, setup_overrides=None, cluster_name="test")
         self.create_cluster_func = None
         self.iterations = 0
 
+    @ignore_log_patterns.setter
```
dcapwell:
```
    @ignore_log_patterns.setter
NameError: name 'ignore_log_patterns' is not defined
```

blah, python docs said this should work!  guess I need a getter... will try that

conftest.py
```diff
@@ -240,11 +241,8 @@ def check_logs_for_errors(dtest_setup):
                     error_str = error.strip()
 
                 if error_str:
-                    logger.error("Unexpected error in {node_name} log, error: \n{error}"
-                                 .format(node_name=node.name, error=error_str))
```
maedhroz, line 244: Was this just creating unnecessary logging noise before?

dcapwell: it was the only thing showing the node_name, so it was needed to know where the logs came from

maedhroz (approved):

PR #195, opened 2022-05-16
Cassandra 17594 attempt 2

dcapwell:


maedhroz (approved):

dcapwell:
test this please

repair_tests/repair_test.py
```diff
@@ -1271,6 +1271,13 @@ def _test_failure_during_repair(self, phase, initiator=False):
             "org.apache.cassandra.repair.SomeRepairFailedException: null"
         ]
 
+        if cluster.version() < '4.0':
+            # CASSANDRA-17466 was not back ported to 3.x, so ignore the error
+            for a in ['java.lang.InterruptedException',
+                      ('java.util.concurrent.CancellationException: '
+                       'Task was cancelled')]:
+                self.fixture_dtest_setup.ignore_log_patterns.append(a)
+
```
ekaterinadimitrova2: Why it was not ported and we ignore it? It is a bug

repair_tests/deprecated_repair_test.py
```diff
@@ -85,6 +85,9 @@ def test_force_repair_range_async_1(self):
                                    Collection<String> hosts, boolean fullRepair,
                                    String... columnFamilies)
         """
+        self.fixture_dtest_setup.ignore_log_patterns = [
+            'Nothing to repair for'
+        ]
```
ekaterinadimitrova2, line 94: Can you provide a bit of info when we see this case and why it is ok to ignore it, please?

dcapwell: This is the log we generate when we skip repair due to no data (or no tables, etc.); this test does *not* write data, so skipping repair is the expected behavior

repair_tests/repair_test.py
dcapwell: Question for @maedhroz, but I think I would answer as the following: repair failed and we have a race condition where we generate an ugly log, the expected behavior of repair isn't impacted, this is a logging issue mostly I believe...

@maedhroz this didn't close the SSTables right?  

If we do wish to backport that ticket, I request we do it in another JIRA

maedhroz: @ekaterinadimitrova2 It's absolutely a bug, but I don't think the burden should be on @dcapwell to fix my lack of a 3.x backport for CASSANDRA-17466 to get this work merged. +1 on a new Jira.

dcapwell: reverted

repair_tests/deprecated_repair_test.py
dcapwell: added docs explaining why; these tests do repairs for a token range, but doesn't put in effort to make sure that there are partitions in that range, so the repair aborts.

The test is checking logging, which is done after we parse arguments and print them... if repair doesn't happen these tests don't really care.

jonmeredith (approved):

PR #199, opened 2022-08-09
Flaky test_rolling_upgrade

bereng:


bereng:
[500](https://app.circleci.com/pipelines/github/bereng/cassandra/721/workflows/fff2c0a7-ea2f-44b1-a42a-d6425984b415/jobs/6503) CI run

It's the same fix as CASSANDRA-17140 so we know it works across all C* versions already.

PR #200, opened 2022-08-17
CASSANDRA-17828: Read/Write/Truncate throw RequestFailure in a race condition with callback timeouts, should return Timeout instead

dcapwell:


cql_test.py
```diff
@@ -796,7 +796,7 @@ def test_truncate_failure(self):
             thrown = True
 
         assert thrown, "No exception has been thrown"
-        assert re.search("Truncate failed on replica /127.0.0.3", str(exception)) is not None
+        assert re.search("Truncate failed on replicas? /127.0.0.3", str(exception)) is not None
```
dcapwell, line 799:
```
>>> import re
>>> re.search('Truncate failed on replica /127.0.0.3', 'Truncate failed on replica /127.0.0.3')
<re.Match object; span=(0, 37), match='Truncate failed on replica /127.0.0.3'>
>>> re.search('Truncate failed on replicas /127.0.0.3', 'Truncate failed on replica /127.0.0.3')
>>> re.search('Truncate failed on replicas? /127.0.0.3', 'Truncate failed on replica /127.0.0.3')
<re.Match object; span=(0, 37), match='Truncate failed on replica /127.0.0.3'>
>>> re.search('Truncate failed on replicas? /127.0.0.3', 'Truncate failed on replicas /127.0.0.3 UNKNOWN')
<re.Match object; span=(0, 38), match='Truncate failed on replicas /127.0.0.3'>
```

PR #202, opened 2022-09-29
Cassandra 14227 trunk

bereng:


client_request_metrics_local_remote_test.py
```diff
@@ -59,7 +59,7 @@ def test_batch_and_slice(self):
         # Run batch test:
         query = 'BEGIN BATCH '
         for i in murmur3_hashes.keys():
-            for y in range(0, 50):
+            for y in range(0, 40):
```
adelapena: Why this change?

bereng: Iirc I was getting 'batch too large` errors but I reverted and they don't seem to show up anymore

ttl_test.py
```diff
@@ -599,7 +603,11 @@ def test_recover_negative_expiration_date_sstables_with_scrub(self):
         node.watch_log_for('Loading new SSTables', timeout=10)
 
         logger.debug("Check that there are no rows present")
-        assert_row_count(session, 'ttl_table', 0)
+        # CASSANDRA-14227 4.2 upwards we have long TTL that can read overflowed rows
+        if self.cluster.version() >= '4.2':
```
adelapena: This should be `5.0` instead of `4.2`.

ttl_test.py
```diff
@@ -365,8 +367,10 @@ def test_expiration_overflow_policy_reject_default_ttl(self):
     def _base_expiration_overflow_policy_test(self, default_ttl, policy):
         """
         Checks that expiration date overflow policy is correctly applied
-        @jira_ticket CASSANDRA-14092
+        @jira_ticket CASSANDRA-14092 and CASSANDRA-14227
         """
+        # Post 5.0 TTL may overflow in 2038 (lgacy) or 2106 C14227
```
adelapena: Typo: legacy

PR #207, opened 2022-11-15
aligned tests with CASSANDRA-18043

smiklosovic:


compaction_test.py
```diff
@@ -293,45 +250,23 @@ def test_compaction_throughput(self):
...
-
-                for x in range(0, 10):
-                    assert_none(session, 'select * from cf where key = ' + str(x))
-
-                node1.flush()
-                cluster.clear()
-                time.sleep(5)
-                cluster.start()
+    @pytest.mark.parametrize("strategy", strategies_without_dtcs)
+    @since('5.0')
```
driftx: What prevents this from running against earlier versions?

compaction_test.py
```diff
@@ -16,7 +16,7 @@
 logger = logging.getLogger(__name__)
 
 strategies = ['LeveledCompactionStrategy', 'SizeTieredCompactionStrategy', 'DateTieredCompactionStrategy']
-
+strategies_without_dtcs = ['LeveledCompactionStrategy', 'SizeTieredCompactionStrategy']
```
jacek-lewandowski: Maybe instead of creating separate lists, and then separate tests for different versions, you would consider creating a function which returns available strategies for a given cluster version, so that the tests would stay mostly version-agnostic?

compaction_test.py
smiklosovic: @driftx nothing, but we do not need to run that, because there is the test above for switching over all of them. This is just a subset to switch over so we do not gain anything, just spending time.

compaction_test.py
smiklosovic: @jacek-lewandowski I can inline this to the annotation directly, it is used in two places only.

jacek-lewandowski (approved):
+1 simply great :)

PR #219, opened 2023-05-25
fix audit log archive command test

krummas:
Patch by marcuse; reviewed by Mick Semb Wever for CASSANDRA-18550


michaelsembwever (approved):
+1 
eyeballed, but not tested

PR #221, opened 2023-06-18
aligned test with CASSANDRA-18555

smiklosovic:


.gitignore
```diff
@@ -11,3 +11,8 @@ upgrade
 html/
 doxygen/doxypy-0.4.2/
 .pytest_cache/
+
+bin/
```
smiklosovic, line 15: this drives me crazy, this should be in `.gitignore` already for a very long time.

topology_test.py
```diff
@@ -443,6 +443,8 @@ def test_crash_during_decommission(self):
         cluster.populate(3).start()
 
         node1, node2 = cluster.nodelist()[0:2]
+        session = self.patient_cql_connection(node1)
+        session.execute("ALTER KEYSPACE system_distributed WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':'2'};")
```
smiklosovic, line 447: this is in other test methods too, without this, we are actually not decommissioning anything because decommission will fail as there will not be enough replicas if replication factor will be 3 and we want to decommission to have just 2. We need to lower this to 2 so by decommissioning we will have still enough.

topology_test.py
```diff
@@ -476,6 +478,8 @@ def test_stop_decommission_too_few_replicas_multi_dc(self):
         @jira_ticket CASSANDRA-12510
         @expected_errors ToolError when # nodes will drop below configured replicas in NTS/SimpleStrategy
         """
+        self.fixture_dtest_setup.ignore_log_patterns = (r'.*Not enough live nodes to maintain replication factor*')
```
smiklosovic, line 485: We need to ignore this in logs, it is a legitimate log to skip as we are testing this scenario but without it, the node would stop suddenly because internal scanning of the logs by dtest framework would falsely evaluate that node to be eligible for termination.

.gitignore
driftx: where is the bin dir coming from though?

topology_test.py
driftx: Comments around this would be useful, it's not obvious what we're doing here.

driftx (approved):

.gitignore
smiklosovic: @driftx there are activation scripts for python env and similar


    $ ls bin/
    activate  activate.csh  activate.fish  activate.nu  activate.ps1  activate_this.py  ccm  pip  pip3  pip-3.10  pip3.10  python  
    python3  python3.10  wheel  wheel3  wheel-3.10  wheel3.10

driftx: That sounds like you are making a virutalenv in the current directory

topology_test.py
smiklosovic: done

.gitignore
smiklosovic: privately agreed we can keep it like it is here

PR #223, opened 2023-06-28
CASSANDRA-18626: Align test_large_compaction_warning with CASSANDRA-18626

adelapena:


compaction_test.py
```diff
@@ -346,6 +347,7 @@ def test_compaction_strategy_switching(self, strategy):
                 time.sleep(5)
                 cluster.start()
 
+    @ported_to_in_jvm('4.0')  # org.apache.cassandra.distributed.test.guardrails.GuardrailPartitionSizeTest
```
driftx: should this be 5.0?

adelapena: Absolutely, it should be 5.0, thanks

PR #224, opened 2023-06-28
Fix native_transport_ssl_test.py::TestNativeTransportSSL::test_connect_to_ssl

jacek-lewandowski:
With the change in Netty, namely https://github.com/netty/netty/pull/13314, it throws NotSslRecordException

### DIFF
--- a/native_transport_ssl_test.py
+++ b/native_transport_ssl_test.py
@@ -40,7 +40,7 @@ class TestNativeTransportSSL(Tester):
         except NoHostAvailable:
             pass
 
-        if cluster.version() >= '4.0':
+        if cluster.version() >= '4.0' and cluster.version() < '5.0':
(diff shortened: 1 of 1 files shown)


native_transport_ssl_test.py
```diff
@@ -40,7 +40,7 @@ def test_connect_to_ssl(self):
         except NoHostAvailable:
             pass
 
-        if cluster.version() >= '4.0':
+        if cluster.version() >= '4.0' and cluster.version() < '5.0':
```
driftx, line 43: If NotSslRecordException is being thrown we should assert that happens in 5.0

jacek-lewandowski: there is "else" block

jacek-lewandowski:
renamed branch https://github.com/apache/cassandra-dtest/pull/226

PR #227, opened 2023-07-18
Permit B as a bytes unit in compaction_test

blambov:
Fixes test failure for the improved `prettyPrintMemory` in CASSANDRA-18397.


driftx (approved):
+1

PR #228, opened 2023-07-20
Fix upgrade_through_versions_test.py::TestUpgrade* tests 

michaelsembwever:
Run generated upgrade_through_versions_test on pytest >7.2.0
pytest-7.2.0 changed how markers were inherited, pytest-dev/pytest#7792

Remove how internode_ssl was changing seeds to append the ssl storage port, it's not needed as the tests always already set enable_legacy_ssl_storage_port to true.

### DIFF
--- a/upgrade_tests/upgrade_through_versions_test.py
+++ b/upgrade_tests/upgrade_through_versions_test.py
@@ -7,12 +7,11 @@ import pprint
 import pytest
 import psutil
 import random
-import re
 import signal
-import subprocess
 import time
 import uuid
 
+from ccmlib.common import get_jdk_version_int
 from collections import defaultdict, namedtuple
 from multiprocessing import Process, Queue
 from queue import Empty, Full
@@ -889,19 +888,31 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
 
 def jdk_compatible_steps(version_metas):
     metas = []
-    java_version = current_env_java_version()
     for version_meta in version_metas:
-        if java_version in version_meta.java_versions:
+        # if you want multi-step upgrades to work with versions that require different jdks
+        #   then define the JAVA<jdk_version>_HOME vars (e.g. JAVA8_HOME)
+        #   ccm detects these variables and changes the jdk when starting/upgrading the node
+        # otherwise the default behaviour is to only do upgrade steps that work with the current jdk
+        javan_home_defined = False
+        for meta_java_version in version_meta.java_versions:
+            javan_home_defined |= 'JAVA{}_HOME'.format(meta_java_version) in os.environ
+        if CURRENT_JAVA_VERSION in version_meta.java_versions or javan_home_defined:
             metas.append(version_meta)
 
     return metas
 
 
 def current_env_java_version():
-    java_command = os.path.join(os.environ['JAVA_HOME'], 'bin', 'java')
-    output = subprocess.check_output([java_command, '-version'], stderr=subprocess.STDOUT).decode('utf-8')
-    return int(re.search('version "(?:1\.)?([0-9]*).*"', output).group(1))
+    # $JAVA_HOME/bin/java takes precedence over any java found in $PATH
+    if 'JAVA_HOME' in os.environ:
+        java_command = os.path.join(os.environ['JAVA_HOME'], 'bin', 'java')
+    else:
+        java_command = 'java'
 
+    return get_jdk_version_int(java_command)
+
+
+CURRENT_JAVA_VERSION = current_env_java_version()
 
 MultiUpgrade = namedtuple('MultiUpgrade', ('name', 'version_metas', 'protocol_version', 'extra_config'))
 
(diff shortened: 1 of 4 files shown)


upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -325,6 +326,8 @@ def fixture_add_additional_log_patterns(self, fixture_dtest_setup):
         )
 
     def prepare(self):
+        if type(self).__name__ == "TestUpgrade":
```
jacek-lewandowski, line 328: I like this change

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -898,6 +887,22 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
...
+    for version_meta in version_metas:
+        if java_version in version_meta.java_versions:
+            metas.append(version_meta)
+
+    return metas
+
+
+def current_env_java_version():
+    java_command = os.path.join(os.environ['JAVA_HOME'], 'bin', 'java')
+    output = subprocess.check_output([java_command, '-version'], stderr=subprocess.STDOUT).decode('utf-8')
```
jacek-lewandowski: this is already implemented:

```
    from ccmlib.common import get_jdk_version_int
    return get_jdk_version_int(java_command)
```

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -898,6 +887,22 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
...
+    java_version = current_env_java_version()
+    for version_meta in version_metas:
+        if java_version in version_meta.java_versions:
+            metas.append(version_meta)
+
+    return metas
+
+
+def current_env_java_version():
+    java_command = os.path.join(os.environ['JAVA_HOME'], 'bin', 'java')
```
jacek-lewandowski: please fail somewhere with a clear message when `JAVA_HOME` is not defined

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -942,18 +947,20 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
...
-                newmeta = oldmeta.clone_with_local_env_version()
-                logger.debug("{} appears applicable to current env. Overriding final test version from {} to {}".format(upgrade.name, oldmeta.version, newmeta.version))
-                metas[-1] = newmeta
-                create_upgrade_class(upgrade.name, [m for m in metas], protocol_version=upgrade.protocol_version, extra_config=upgrade.extra_config)
-        else:
-            create_upgrade_class(upgrade.name, [m for m in metas], protocol_version=upgrade.protocol_version, extra_config=upgrade.extra_config)
+            # replace matching meta with current version
+            for idx, meta in enumerate(metas):
+                if meta.matches_current_env_version_family:
+                    java_version = current_env_java_version()
```
jacek-lewandowski: we should get that java version once and do not do that in every iteration in a loop

jacek-lewandowski (commented):
You should also make sure to skip the entire test if say we run with j17 and the upgrade path ends up with a single version

upgrade_tests/upgrade_through_versions_test.py
michaelsembwever: it should only called once, i was missing a `break`

upgrade_tests/upgrade_through_versions_test.py
michaelsembwever: not needed, see update.

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -942,17 +957,21 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
...
-            if metas[-1].matches_current_env_version_family:
-                # looks like this test should actually run in the current env, so let's set the final version to match the env exactly
-                oldmeta = metas[-1]
-                newmeta = oldmeta.clone_with_local_env_version()
-                logger.debug("{} appears applicable to current env. Overriding final test version from {} to {}".format(upgrade.name, oldmeta.version, newmeta.version))
-                metas[-1] = newmeta
-                create_upgrade_class(upgrade.name, [m for m in metas], protocol_version=upgrade.protocol_version, extra_config=upgrade.extra_config)
-        else:
+        # even for RUN_STATIC_UPGRADE_MATRIX we only test upgrade paths jdk compatible with the end "indev_" version (or any JAVA<jdk_version>_HOME defined)
+        metas = jdk_compatible_steps(upgrade.version_metas)
```
jacek-lewandowski, line 962: you are still calling `current_env_java_version()` in a loop

jacek-lewandowski (approved):

upgrade_tests/upgrade_through_versions_test.py
michaelsembwever: oh, your talking about the outer loop `for upgrade in MULTI_UPGRADES:`

PR #237, opened 2023-10-13
CASSANDRA-18742 followup

smiklosovic:
Before 18742 was in, it was possible to specify "sstable_compression" as well as "class" in compression parameters. This was true from times of 3.0 to trunk.

In 2.x, "class" was not there yet, only "sstable_compression".

When I was implementing 18742, I modified dtest.py to get rid of "sstable_compression" so it will work only with "class".

The problem with this is that when there are upgrade tests from 2.0 to 3.0 / 3.11, we create a table on 2.0 with "class" but it is not recognized yet.

The fix is to create a table with "sstable_compression" if the node is of version 2.x

### DIFF
--- a/sstable_generation_loading_test.py
+++ b/sstable_generation_loading_test.py
@@ -10,7 +10,7 @@ import logging
 from ccmlib import common as ccmcommon
 from ccmlib.node import ToolError
 
-from dtest import Tester, create_ks, create_cf, mk_bman_path, MAJOR_VERSION_4
+from dtest import Tester, create_ks, create_cf, mk_bman_path, MAJOR_VERSION_4, MAJOR_VERSION_5
 from tools.assertions import assert_all, assert_none, assert_one
 
 since = pytest.mark.since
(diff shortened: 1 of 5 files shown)


sstable_generation_loading_test.py
```diff
@@ -35,10 +35,10 @@ def compact(self):
         return self.fixture_dtest_setup.cluster.version() < MAJOR_VERSION_4 and self.test_compact
 
     def create_schema(self, session, ks, compression):
+        legacy_compression_class = self.fixture_dtest_setup.cluster.version() < MAJOR_VERSION_3
```
Mmuzaf: The cassandra-18742 patch is fixed as of 5.0, so the obsolete param should work for versions < 5.0, right? I wonder if we can fix this condition...

smiklosovic: @Mmuzaf  yes, you are technically right, it will work with `sstable_compression` until 5.0 but I am thinking .... why not to use non-deprecated param already for all branches which support it?

Mmuzaf: yes, but in this case it is not clear from which version the parameter was removed :-(

Mmuzaf (approved):

PR #241, opened 2023-11-15
Debug/slow upgrade tests

dcapwell:


dtest.py
```diff
@@ -249,6 +249,12 @@ def set_dtest_setup_on_function(self, fixture_dtest_setup):
         self.dtest_config = fixture_dtest_setup.dtest_config
         return None
 
+    @pytest.fixture(autouse=True)
+    def cleanup_connections(self):
+        yield None
```
aratno, line 254: Can you add a comment about why this yield is necessary?

aratno (approved):

dtest.py
dcapwell: this is documented when you look up `fixture`, its a core way that this works... w/e is before `yield` is setup, w/e is after `yield` is cleanup.

aratno: Ah right on - no extra doc necessary then

dcapwell: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

PR #245, opened 2023-11-27
Add version guard around call to initializecms also in upgrade_tests/…

beobal:
…storage_engine_upgrade_test.py

 patch by Mick Semb Wever; reviewed by XXX for CASSANDRA-19099

### DIFF
--- a/upgrade_tests/storage_engine_upgrade_test.py
+++ b/upgrade_tests/storage_engine_upgrade_test.py
@@ -69,7 +69,7 @@ class TestStorageEngineUpgrade(Tester):
         node1.set_install_dir(install_dir=self.fixture_dtest_setup.default_install_dir)
         self.install_legacy_parsing(node1)
         node1.start(wait_for_binary_proto=True)
-        if self.dtest_config.cassandra_version_from_build >= MAJOR_VERSION_5:
+        if node1.get_cassandra_version() >= '5.1':
             node1.nodetool("initializecms")
         if self.fixture_dtest_setup.bootstrap:
             cluster.set_install_dir(install_dir=self.fixture_dtest_setup.default_install_dir)


upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -69,7 +69,7 @@ def _do_upgrade(self, login_keyspace=True):
         node1.set_install_dir(install_dir=self.fixture_dtest_setup.default_install_dir)
         self.install_legacy_parsing(node1)
         node1.start(wait_for_binary_proto=True)
-        if self.dtest_config.cassandra_version_from_build >= MAJOR_VERSION_5:
+        if node1.get_cassandra_version() >= '5.1':
```
beobal, line 72: This seems to be the cause of all the failures in CASSANDRA-19099, we missed this when bumping TCM from 5.0 to 5.1.

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -497,7 +497,7 @@ def upgrade_scenario(self, populate=True, create_schema=True, rolling=False, aft
                 self._check_counters()
                 self._check_select_count()
 
-        if self.cluster.version() >= '5.1':
+        if self.cluster.nodelist()[0].get_cassandra_version() >= '5.1':
```
beobal: This I don't believe is correct. I can see no valid reason why the cluster would be on 5.1 but node1 would not be & in fact that would constitute a failure at this point in the test.

michaelsembwever: we can do forward upgrade tests.
so the cluster is configured with 5.0, and then one node is upgraded to 5.1


and the guarded line is specific to that node 🤷

michaelsembwever: my bad – this method upgrades all nodes and the cluster first.

upgrade_tests/storage_engine_upgrade_test.py
beobal: 👍

beobal (commented):
+1

michaelsembwever:
Committed https://github.com/apache/cassandra-dtest/commit/eb518f3ca6f22753346dfb57c94e9935611c6a53

PR #246, opened 2023-11-28
CASSANDRA-19103 fix counter test

smiklosovic:


counter_test.py
```diff
@@ -114,8 +114,11 @@ def test_counter_leader_with_partial_view(self):
 
         # Now stop the node and restart but first install a rule to slow down how fast node 2 will update the list
         # nodes that are alive
-        nodes[1].stop(wait=True, wait_other_notice=False)
-        nodes[1].update_startup_byteman_script(mk_bman_path('gossip_alive_callback_sleep.btm'))
+        nodes[1].stop(wait=True, wait_other_notice=True)
```
smiklosovic, line 117: this is here to tell nodes which are still up that the second node went offline. Without that, it will still think that the second node is up.

counter_test.py
```diff
@@ -114,8 +114,11 @@ def test_counter_leader_with_partial_view(self):
...
+
+        if cluster.version() < '5.1':
+            nodes[1].update_startup_byteman_script(mk_bman_path('gossip_alive_callback_sleep.btm'))
+
         nodes[1].start(no_wait=True, wait_other_notice=False)
```
smiklosovic, line 122: here we do not wait for other nodes to notice the second one is up, it will be detected on their own when the logic is correct. Until they see the second node JOINED after we started it, they will never select that on their own.

byteman/4.0/election_counter_leader_favor_node2.btm
```diff
@@ -7,8 +7,8 @@ RULE election counter leader cheat
 CLASS org.apache.cassandra.service.StorageProxy
 METHOD findSuitableEndpoint
 AT EXIT
-BIND isthere:boolean = $localEndpoints.contains(org.apache.cassandra.locator.InetAddressAndPort.getByName("127.0.0.2"));
+BIND isthere:boolean = $localReplicas.stream().anyMatch(r -> { try { return r.endpoint().equals(org.apache.cassandra.locator.InetAddressAndPort.getByName("127.0.0.2")); } catch (Throwable t) { return false; } });
```
smiklosovic, line 10: `localEndpoints` are called `localReplicas` in 4.0+ and it is a collection of `Replica`s, not `InetAddress(WithPort)`

PR #249, opened 2023-12-17
CASSANDRA-14572 Add system_metrics keyspace to default keyspaces list

Mmuzaf:


cqlsh_tests/test_cqlsh.py
```diff
@@ -1040,7 +1040,7 @@ def test_default_keyspaces_exist(self):
         node1, = self.cluster.nodelist()
 
         # Describe keyspaces
-        expected_keyspaces = ['system_schema', 'system', 'system_traces', 'system_views',
+        expected_keyspaces = ['system_schema', 'system', 'system_traces', 'system_views', 'system_metrics',
```
smiklosovic: @Mmuzaf this will not work because system_metrics will be only on trunk and this test will be executed for 4.0+ so when 4.0 will be built, this one will not be there. You need to do it like below for system_cluster_metadata case:

        if self.cluster.version() >= LooseVersion('5.1'):
            expected_keyspaces.append('system_cluster_metadata')
            expected_keyspaces.append('system_metrics')

Mmuzaf: Makes sense, fixed.

PR #254, opened 2024-02-23
Fix timeouts in pytest.ini to not take preference over test decorations

ekaterinadimitrova2:
CASSANDRA-19409


ekaterinadimitrova2:
Merged manually

PR #258, opened 2024-03-15
CASSANDRA-19474 fix tests to cope with cassandra_latest.yaml

smiklosovic:


cqlsh_tests/test_cqlsh.py
```diff
@@ -127,6 +127,21 @@ def tearDown(self):
             os.unlink(self.tempfile.name)
         super(TestCqlsh, self).tearDown()
 
+
+    def latest_config(self):
```
driftx: We'll probably need this again, so we could put in dtest.py

PR #259, opened 2024-03-22
Remove error test already covered by unit tests for CASSANDRA-19341

blerer:


json_test.py
```diff
@@ -1238,38 +1238,6 @@ def test_simple_schema(self):
         """
         run_func_docstring(tester=self, test_func=self.test_simple_schema)
 
-    def test_pkey_requirement(self):
```
ekaterinadimitrova2, line 1241: Can you add to the commit message which unit tests cover it, for clarity? Thanks

cql_test.py
```diff
@@ -1219,12 +1219,6 @@ def _assert_logs_slow_queries_with_skinny_table(node, session):
         TestCQLSlowQuery._assert_logs(node, session, table,
                                       query="SELECT * FROM {} WHERE k >= 0 AND k < 100 ALLOW FILTERING",
                                       logged_query="SELECT \* FROM ks.{} WHERE k >= 0 AND k < 100")
-        TestCQLSlowQuery._assert_logs(node, session, table,
-                                      query="SELECT * FROM {} WHERE k <= 100 AND k > 0 ALLOW FILTERING",
-                                      logged_query="SELECT \* FROM ks.{} WHERE k > 0 AND k <= 100")
-        TestCQLSlowQuery._assert_logs(node, session, table,
-                                      query="SELECT * FROM {} WHERE k < 100 AND k >= 0 ALLOW FILTERING",
-                                      logged_query="SELECT \* FROM ks.{} WHERE k >= 0 AND k < 100")
```
ekaterinadimitrova2, line 1227: So technically, now we preserve the restrictions order, and we can fix the two `logged_query.`?
I am wondering why you decided to remove the two assertions instead. This test covered some wrong logging reported in CASSANDRA-15503. Maybe it is better to keep those assertions to ensure no future regressions?

json_test.py
ekaterinadimitrova2: I saw you added respective unit test in the cassandra repo. We should either backport that test to previous versions too or we should keep this one up to 5.0.
Something like this - `@since("2.2", max_version="4.X")`. and port the cassandra test only to 5.0 maybe?
We can say up to 5.0 but then it won't test probably patch versions of 5.0, so 4.X sounds most reasonable to me.

cql_test.py
ekaterinadimitrova2, line 1227:
```
        if node.cluster.version() >= '5.1':
            TestCQLSlowQuery._assert_logs(node, session, table,
                                          query="SELECT * FROM {} WHERE k <= 100 AND k > 0 ALLOW FILTERING",
                                          logged_query="SELECT \* FROM ks.{} WHERE k <= 100 AND k > 0")
            TestCQLSlowQuery._assert_logs(node, session, table,
                                          query="SELECT * FROM {} WHERE k < 100 AND k >= 0 ALLOW FILTERING",
                                          logged_query="SELECT \* FROM ks.{} WHERE k < 100 AND k >= 0")
        else:
            TestCQLSlowQuery._assert_logs(node, session, table,
                                          query="SELECT * FROM {} WHERE k <= 100 AND k > 0 ALLOW FILTERING",
                                          logged_query="SELECT \* FROM ks.{} WHERE k > 0 AND k <= 100")
            TestCQLSlowQuery._assert_logs(node, session, table,
                                          query="SELECT * FROM {} WHERE k < 100 AND k >= 0 ALLOW FILTERING",
                                          logged_query="SELECT \* FROM ks.{} WHERE k >= 0 AND k < 100")
```

This passes for me both with your patch and plain 5.0 branch.

ekaterinadimitrova2 (commented):
Left a few tiny comments around versioning as these tests are run for all branches

ekaterinadimitrova2 (approved):

maedhroz (approved):
LGTM, assuming outstanding comments are resolved one way or another (even if that's just correcting my ignorance)

PR #262, opened 2024-05-23
Fix dtest failing after CASSANDRA-19592

beobal:


beobal:
+1

beobal:
Committed with a slight fix to maintain compatibility with prior Cassandra versions in 13b265647b510b5caaa1b3dd978b165e4e512214

PR #269, opened 2024-10-06
Align dtest.py for CASSANDRA-19671

smiklosovic:
patch by Stefan Miklosovic; reviewed by Brad Schoening for CASSANDRA-19671

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -414,7 +414,7 @@ def data_size(node, ks, cf):
     """
     Return the size in bytes for given table in a node.
     This gets the size from nodetool tablestats output.
-    This might brake if the format of nodetool tablestats change
+    This might break if the format of nodetool tablestats change
     as it is looking for specific text "Space used (total)" in output.
     @param node: Node in which table size to be checked for
     @param ks: Keyspace name for the table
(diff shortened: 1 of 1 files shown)


dtest.py
```diff
@@ -426,11 +426,12 @@ def data_size(node, ks, cf):
     regex = re.compile(r'[\t]')
```
bschoening, line 426: There is a typo in the method comment, 'brake' should be 'break'.

dtest.py
```diff
@@ -426,11 +426,12 @@ def data_size(node, ks, cf):
...
     stats_lines = [regex.sub("", s) for s in tablestats.split('\n')
                   if regex.sub("", s).startswith('Space used (total)')]
-    if not len(stats_lines) == 1:
-        msg = ('Expected output from `nodetool tablestats` to contain exactly 1 '
+    if len(stats_lines) > 2:
+        msg = ('Expected output from `nodetool tablestats` to contain at most 2 '
                'line starting with "Space used (total)". Found:\n') + tablestats
```
bschoening: Msg should be plural, 'lines starting with...'

bschoening (changes requested):
+1 on the code change.  There are two small typo in the method comment, 'brake' should be 'break', and 'line' vs 'lines'.

PR #275, opened 2025-04-09
For upgrade tests correctly use cassandra_version over version, and set protocol_version to highest source version supports

michaelsembwever:
This PR…
    1.  Fix places where we need to be using `cassandra_version()` (instead of just `version()`), now that we support other cluster types.
    2.  for upgrade tests default the protocol_version to the max of the source/starting version in the upgrade path -- the assumption here is that from.max_proto_v will always be within the to.min_proto_v to to.max_proto_v range.
    3.  for upgrade tests default the endpoint_snitch yaml property to SimpleSnitch. this relates to HCD-129
    4.  after upgrading a node, check and prune any unsupported config options
    5.  when creating a cluster (`dtest_setup.create_ccm_cluster(..)`) use the correct cluster type subclass
    6.  when upgrading, check (and change if necessary) the cluster class type and the node class type (and copy config files into new config locations)

### DIFF
--- a/dtest_setup.py
+++ b/dtest_setup.py
@@ -17,7 +17,8 @@ from cassandra.cluster import NoHostAvailable
 from cassandra.cluster import EXEC_PROFILE_DEFAULT
 from cassandra.policies import WhiteListRoundRobinPolicy
 from ccmlib.common import is_win
-from ccmlib.cluster import Cluster
+from ccmlib.cluster import Cluster, NodeError
+from ccmlib.extension import get_cluster_class
 
 from dtest import (get_ip_from_node, make_execution_profile, get_auth_provider, get_port_from_node,
                    get_eager_protocol_version, hack_legacy_parsing)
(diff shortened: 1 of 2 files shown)


dtest_setup.py
```diff
@@ -439,7 +440,7 @@ def init_default_config(self):
         phi_values = {'phi_convict_threshold': 5}
 
         # enable read time tracking of repaired data between replicas by default
-        if self.cluster.version() >= '4':
+        if self.cluster.cassandra_version() >= '4':
```
bereng, line 443: Don't you also need to make that change in L98?

michaelsembwever: L98 is
```
        if self.cluster.version() < LooseVersion('3.11.13'):
```
in `def install_nodetool_legacy_parsing`

i don't think so…  what were your thoughts here ?

bereng: I am obviously missing what is the difference of comparing against a version number and a `LooseVersion`. Aren't you trying to change behavior depending on the 'embedded' underlying C* versions instead of the 'product' version?

michaelsembwever: Yes, you're quite right.  Call me blind.

bereng:
+1 despite being Draft conditioned to good CI

michaelsembwever:
committed https://github.com/apache/cassandra-dtest/commit/4f6f06c66e4247072bda6587db52e4dc24dfb784

PR #281, opened 2025-07-31
CASSANDRA-20389: Fix expected error message for too long table name

k-rus:
https://github.com/apache/cassandra/pull/4038 changes the error message, thus the expected error message needs to be updated in tests.

Currently it's changed for all versions from 4.0. I am not sure if it will be necessary to be specific for 4.0.x.

### DIFF
--- a/cqlsh_tests/test_cqlsh.py
+++ b/cqlsh_tests/test_cqlsh.py
@@ -599,7 +599,7 @@ UPDATE varcharmaptable SET varcharvarintmap['Vitrum edere possum, mihi non nocet
             err = node1.run_cqlsh(cmds=cmd, cqlsh_options=options).stderr
 
         if self.cluster.version() >= LooseVersion('4.0'):
-            assert "Keyspace name must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters (got 'ä')" in err
+            assert 'Keyspace name must not be empty and must contain alphanumeric or underscore characters only (got "ä")' in err
         else:
             assert '"ä" is not a valid keyspace name' in err
 


cqlsh_tests/test_cqlsh.py
```diff
@@ -599,7 +599,7 @@ def test_unicode_invalid_request_error(self):
             err = node1.run_cqlsh(cmds=cmd, cqlsh_options=options).stderr
 
         if self.cluster.version() >= LooseVersion('4.0'):
-            assert "Keyspace name must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters (got 'ä')" in err
+            assert 'Table name must not be empty or not contain non-alphanumeric-underscore characters (got "ä")' in err
```
netudima: Should we have a keyspace name here?
```suggestion
            assert 'Keyspace name must not be empty or not contain non-alphanumeric-underscore characters (got "ä")' in err
```

k-rus: Good catch! Sorry that I forgot to check this PR.

netudima (approved):

cqlsh_tests/test_cqlsh.py
```diff
@@ -599,7 +599,7 @@ def test_unicode_invalid_request_error(self):
             err = node1.run_cqlsh(cmds=cmd, cqlsh_options=options).stderr
 
         if self.cluster.version() >= LooseVersion('4.0'):
-            assert "Keyspace name must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters (got 'ä')" in err
+            assert 'Keyspace name must not be empty or not contain non-alphanumeric-underscore characters (got "ä")' in err
```
pkolaczk: Too many "not"?
I'm not a native speaker, but this parses to me like:

`Keyspace name must not: (be empty) or (not contain non-alphanumeric-underscore characters) (got "ä")`

So for the second case it reduces to:
`Keyspace name must not not contain non-alphanumeric-underscore characters (got "ä")`

I think that "not" before "contain" is unneeded.
It was correct before.

pkolaczk: If the original wasn't unambiguous enough, then maybe this would work better:

`Keyspace name must not be empty and must not contain non-alphanumeric-underscore characters (got "ä")` 

?

pkolaczk (changes requested):
See the subcomment on the message

cqlsh_tests/test_cqlsh.py
k-rus: @pkolaczk your comment is not to this PR, but to C* PR, where the error message is introduced :) I will update there, which will fix it here too. Thank you for noticing the wording and proposing the improvement.

Maxwell-Guo:
+1

Maxwell-Guo (approved):

cqlsh_tests/test_cqlsh.py
```diff
@@ -599,7 +599,7 @@ def test_unicode_invalid_request_error(self):
             err = node1.run_cqlsh(cmds=cmd, cqlsh_options=options).stderr
 
         if self.cluster.version() >= LooseVersion('4.0'):
-            assert "Keyspace name must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters (got 'ä')" in err
+            assert 'Keyspace name must not be empty and must contain alphanumeric or underscore characters (got "ä")' in err
```
pkolaczk: This is much better.
Minor nit: add "only" at the end:

`Keyspace name must not be empty and must contain alphanumeric or underscore characters only`

pkolaczk (approved):
LGTM. Just a minor nit, no more reviews needed.

PR #284, opened 2026-01-28
Add JDK21 support for CASSANDRA-18831

jmckenzie-dev:
- Fix uninitialized java_version in upgrade_through_versions_test.py
- Add JDK21 to indev_trunk in upgrade_manifest.py
- Skip mantissa including json_test.py testcases on JDK19+
- Make byteman units explicit for sleeps

### DIFF
--- a/byteman/4.0/repair_validation_sleep.btm
+++ b/byteman/4.0/repair_validation_sleep.btm
@@ -9,5 +9,5 @@ AT ENTRY
 IF NOT flagged("done")
 DO
    flag("done");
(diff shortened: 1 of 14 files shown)


michaelsembwever (approved):
I assume this should be merged before in-tree, i.e. next.

jmckenzie-dev:
Yep. And I _believe_ merging this should be non-disruptive to existing branches. You agree? If so, let's pull the trigger here so we can move on the other.

PR #286, opened 2026-02-23
CASSANDRA-21182 - Fix flaky DTest: largecolumn_test.TestLargeColumn

samueldlightfoot:
Force GC before direct memory measurements for ZGC compatibility

Generational ZGC (JDK 21 default) only processes PhantomReferences during major collections, which may not trigger during short test runs. This causes DirectByteBuffer Cleaners to not fire, leaving ~63MB of stale direct memory that inflates TOTAL_CAPACITY and fails the assertion.

Force System.gc() via nodetool before each measurement to ensure a major collection runs and reference processing completes.

CASSANDRA-21182

### DIFF
--- a/largecolumn_test.py
+++ b/largecolumn_test.py
@@ -1,6 +1,8 @@
 import pytest
 import re
 import logging
+import subprocess
+import time
 
 from dtest import Tester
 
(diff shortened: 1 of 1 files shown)


largecolumn_test.py
```diff
@@ -69,11 +71,18 @@ def test_cleanup(self):
...
+        # which may not run during short tests. Force a GC; sleep lets the async Cleaner
+        # thread finish decrementing TOTAL_CAPACITY before we measure.
+        subprocess.check_call(['jcmd', str(node1.pid), 'GC.run'])
+        time.sleep(1)
         after1stLargeStress = self.directbytes(node1)
         logger.info("After 1st large column stress, direct memory: {0}".format(after1stLargeStress))
 
         # Now run the full stack to see how much memory is allocated for the second "large" columns request
         self.stress_with_col_size(cluster, node1, LARGE_COLUMN_SIZE)
+        subprocess.check_call(['jcmd', str(node1.pid), 'GC.run'])
```
netudima: What about adding a limited retry loop (sleep, check direct bytes, repeat) to avoid flakiness?.. Sometimes CI is really slow, and 1 second might not be enough.

samueldlightfoot: Good point - pushed with fixed wait retry to verify memory convergence.

netudima (approved):

largecolumn_test.py
samueldlightfoot: Agreed to increase timeout from 1s to 3s due to no simple way to ensure the cleaner has finished

PR #296, opened 2026-07-13
(CASSANDRA-21316) Fix disk_balance_test flakiness under Paxos v2 (dtest-latest)

michaelsembwever:
- test_disk_balance_replace_*: raise replacement-node startup timeout so paxos repair for bootstrap can retry (~2 min/attempt observed) instead of skipping the repair, keeping Paxos v2 topology-repair coverage
 - test_disk_balance_decommission: ignore the transient 'Error during paxos repair' log line; the repair still runs and a genuine failure still aborts the decommission and fails the test

### DIFF
--- a/disk_balance_test.py
+++ b/disk_balance_test.py
@@ -97,8 +97,11 @@ class TestDiskBalance(Tester):
                      jmx_port='7500', remote_debug_port='0', initial_token=None,
                      binary_interface=(node5_address, 9042))
         self.cluster.add(node5, False, data_center="dc1")
+        # Paxos v2 (dtest-latest) paxos repair (while bootstrapping replacement node) can time out and retry
+        paxos_variant = node1.get_conf_option('paxos_variant')
+        startup_timeout = 600 if paxos_variant is not None and paxos_variant.startswith('v2') else 180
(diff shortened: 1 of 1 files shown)


disk_balance_test.py
```diff
@@ -108,6 +111,10 @@ def test_disk_balance_decommission(self):
         cluster = self.cluster
         if self.dtest_config.use_vnodes:
             cluster.set_configuration_options(values={'num_tokens': 256})
+        # Paxos v2 (dtest-latest) paxos repair (while bootstrapping replacement node) can time out and retry
+        paxos_variant = node1.get_conf_option('paxos_variant')
```
netudima, line 117: node1 variable is assigned after this line, should we move the declaration after assignment?

disk_balance_test.py
```diff
@@ -108,6 +111,10 @@ def test_disk_balance_decommission(self):
         cluster = self.cluster
         if self.dtest_config.use_vnodes:
             cluster.set_configuration_options(values={'num_tokens': 256})
+        # Paxos v2 (dtest-latest) paxos repair (while bootstrapping replacement node) can time out and retry
```
netudima: nit: "bootstrapping replacement node" - there is no replacement in this test, need to adjust the comment

netudima (approved):

